PR #3, opened 2022-11-29
Setup test fixtures and add example test for `aggregate_temporal_period`

LukeWeidenwalker:
I've let myself be inspired by the testsuite of https://github.com/makepath/xarray-spatial and have added a couple of test functions and fixtures in that spirit! I've also added an example test for aggregate_temporal_period that uses these.

### DIFF
--- a/tests/mockdata.py
+++ b/tests/mockdata.py
@@ -1,6 +1,5 @@
 import logging
 
-import dask.array as da
 import numpy as np
 import pandas as pd
 import xarray as xr
@@ -52,6 +51,8 @@ def create_fake_rastercube(
     raster_cube.rio.write_crs(spatial_extent.crs, inplace=True)
 
     if "dask" in backend:
+        import dask.array as da
+
         raster_cube.data = da.from_array(raster_cube.data, chunks=chunks)
 
     return raster_cube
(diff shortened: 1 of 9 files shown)


tests/mockdata.py
```diff
@@ -27,36 +25,34 @@ def generate_fake_rastercube(
...
-    da.random.seed(int(str(abs(hash(seed)))[:9]))
-    _data = da.random.random(tuple([len(v) for _, v in coords.items()]))
+    coords = {"x": x_coords, "y": y_coords, "t": t_coords, "bands": bands}
 
-    data = xr.DataArray(
-        data=_data,
+    raster_cube = xr.DataArray(
+        data=data,
         coords=coords,
         attrs={"crs": spatial_extent.crs},
```
SerRichard, line 49: keep meaning to mention, but we probably shoudn't set data crs values to actualy CRS objects, either the wkt string or just the epsg code is enough and usable, the CRS object will cause an exception when trying to write the data to a netcdf or a gtiff, and imo it's a bit annoying to reset the values to string when wanting to write. Wdyut? If we agree might be worth actually changing this upstream in the parser too so it's clearer throughout.

LukeWeidenwalker: Ahh, didn't realise that was a problem! That's a fair point, we can just use the CRS object during validation, but then store the wkt string going forward!

PR #5, opened 2022-12-01
Wrap process registry in a class

LukeWeidenwalker:
Closes [EODCAPI-81](https://eodc.atlassian.net/browse/EODCAPI-81).

This is to formalise the ideas of process_id trimming and process aliases into a single class that can be imported and used elsewhere.


openeo_processes_dask/core.py
```diff
@@ -67,12 +58,79 @@ def wrapper(*args, parameters: Optional[dict[str]] = None, **kwargs):
...
+    """
+
+    def __init__(self, *args, **kwargs):
+        self.store = dict()  # type: dict[str, Callable]
+        self.aliases = dict()  # type: dict[str, str]
+
+        self.update(dict(*args, **kwargs))  # use the free update to set keys
+
+    def __getitem__(self, key) -> Callable:
+        return self.store[self._keytransform(key)]
```
SerRichard, line 74: If I enter the name of an alias here, it would KeyError? If we are allowing aliases, would be cool to support them here?

SerRichard (commented):
Lgtm, one comment!

openeo_processes_dask/core.py
SerRichard: nice!

SerRichard (approved):

Issue #15, opened 2022-12-06
openEO exceptions 


clausmichele:
Looking at the current implementation of the openEO expections I was thinking about possible improvements. Currently, a new class is defined for every exception.
I tested successfully a dynamic way of generating an exception class, could we discuss about it @LukeWeidenwalker?
I could create a PR soon!

It would avoid to create a new class for every new exception, we would just have to add it to the dictionary.
It would also avoid to have to import all the exceptions like:
`from openeo_processes_dask.exceptions import ProcessParameterMissing`
Additionally, it can also return a class with a custom message if it's not already in the dict.

```python
from typing import Optional

openeo_exceptions_dict_mapping = {
    "DimensionNotAvailable": "A dimension with the specified name does not exist.",
    "DimensionLabelCountMismatch": "The number of dimension labels exceeds one, which requires a reducer.",
    "ArrayElementParameterConflict" : "The process `array_element` only allows that either the `index` or the `labels` parameter is set.",
    "ArrayElementParameterMissing" : "The array has no element with the specified index or label.",
    "ArrayNotLabeled" : "The array is not a labeled array, but the `label` parameter is set. Use the `index` instead.",
    "ArrayElementNotAvailable" : "The array has no element with the specified index or label.",
    "TooManyDimensions": "The number of dimensions must be reduced to three for `aggregate_spatial`."
}

def exception_constructor(self, msg):
    self.message = msg
    
def exception_to_string(self):
    return self.message

def OpenEOException(openeo_exception: str = None,
                      message: Optional[str] = None):
    exception_class = type(openeo_exception, (Exception, ), {
    "__init__": exception_constructor,
    "__str__": exception_to_string
    })
    if message is not None:
        return exception_class(openeo_exception,message)
    else:
        return exception_class(openeo_exceptions_dict_mapping[openeo_exception])
```

The usage would be
```python

from openeo_processes_dask.exceptions import OpenEOException

raise OpenEOException("DimensionNotAvailable")
```

LukeWeidenwalker:
Thanks for the suggestion - I just tried this, and it seems that this change would prevent catching these exceptions with except.

```
try:
    raise OpenEOException("DimensionNotAvailable")
except OpenEOException:  # same with `except OpenEOException("DimensionNotAvailable"):`
    print("caught")
```
gives a TypeError.

Why do you think this change is necessary?

clausmichele:
[...]

PR #19, opened 2022-12-12
Updates to exceptions

LukeWeidenwalker:
Addresses #15 in part.

### DIFF
--- a/openeo_processes_dask/process_implementations/cubes/general.py
+++ b/openeo_processes_dask/process_implementations/cubes/general.py
@@ -17,7 +17,7 @@ def drop_dimension(data: RasterCube, name: str) -> RasterCube:
         )
     if len(data[name]) > 1:
         raise DimensionLabelCountMismatch(
-            "The number of dimension labels exceeds one, which requires a reducer."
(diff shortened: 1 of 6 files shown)


openeo_processes_dask/process_implementations/cubes/general.py
```diff
@@ -12,9 +12,13 @@
...
 def drop_dimension(data: RasterCube, name: str) -> RasterCube:
     if name not in data.dims:
-        raise DimensionNotAvailable()
+        raise DimensionNotAvailable(
+            f"Provided dimension ({name}) not found in data.dims: {data.dims}"
+        )
     if len(data[name]) > 1:
-        raise DimensionLabelCountMismatch()
+        raise DimensionLabelCountMismatch(
+            "The number of dimension labels exceeds one, which requires a reducer."
```
clausmichele, line 20: We could make this error message more informative as well, like:
f"The number of dimension labels exceeds one, which requires a reducer. Dimension ({name}) has {len(data[name])} labels."

clausmichele (changes requested):
See my comment on the drop_dimension error message.

PR #21, opened 2022-12-19
refactor: Avoid passing datacubes into all math/array processes

LukeWeidenwalker:
~~Some preliminary results of me playing around with the `xarray.reduce` API. Think our callbacks functions aren't quite right.~~

Here's a full changeset for following the API spec more closely. I think this is important for the following reasons:
- It prevents the sub processes from manipulating datacubes, (i.e. potentially losing `attrs` and needing to call `keep_attrs` all the time, changing other metadata). Thereby we only need to test correct metadata handling in the parent function, rather than across all these child processes (or not at all, which is even worse)
- If done right, this might opens back up the possibility of installing just the numpy processes without dask after all. 
- Equally important: It makes our codebase more predictable. I find it totally confusing if the implementation for a process that in the spec takes `array`, really expects a datacube.
- This also gives us a nice place for testing the dask-friendliness of the child processes, i.e. computing numpy results first, and then comparing against dask results. 

I've tested performance on a subset of those changes, and there's no difference. All the numpy functions used in these processes have an equivalent implementation in dask. Calling these functions on dask objects will use dask's function dispatch to correctly route to the dasky implementation.

A follow-on change will be to update `apply`, `apply_dimension`, `merge_cubes` and the other parent processes to do it like this too.

Let me know what you think!


clausmichele:
Hi Lukas. Even if the reduce operations were not implemented exactly as specified in the openEO API, so working directly on datacubes instead of datacubes (xarray.DataArray), shouldn't be an issue if the result was the same.
Anyway, it would be good to have a performance benchmark between the two implementations before merging.

LukeWeidenwalker:
I've done some adhoc benchmarking this afternoon, `.reduce` seems to be exactly as performant as going through the methods on dataarray - which isn't surprising, because under the hood, these also use the `.reduce` API ;) 

Have also done the same testing for `apply_ufunc`, with similar results. Will use that to replace the implementation of `apply`.

clausmichele:
Thanks for digging into this, it will really help to align more with the openEO API!

clausmichele (approved):

openeo_processes_dask/core.py
```diff
@@ -30,16 +42,19 @@ def wrapper(*args, parameters: Optional[dict[str]] = None, **kwargs):
...
         resolved_kwargs = {}
-        for k, v in kwargs.items():
-            if isinstance(v, ParameterReference):
-                if v.from_parameter in parameters:
-                    resolved_kwargs[k] = parameters[v.from_parameter]
+        for k, arg in kwargs.items():
+            if isinstance(arg, ParameterReference):
+                if arg.from_parameter in named_parameters:
+                    resolved_kwargs[k] = named_parameters[arg.from_parameter]
+                elif arg.from_parameter in positional_parameters:
```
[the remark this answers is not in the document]
LukeWeidenwalker: Ah, yes, because it should just pass over, not enter the else, where an error is raised!

openeo_processes_dask/process_implementations/cubes/apply.py
```diff
@@ -9,8 +11,18 @@
 def apply(
     data: RasterCube, process: Callable, context: Optional[dict] = None, **kwargs
 ) -> RasterCube:
-    parameters = {"x": data, "context": context}
```
SerRichard, line 12: illuminating what context and positional_parameters are used for in a docstring could be pretty useful here, same for context in reduce

SerRichard (commented):
Couple of high level comments

openeo_processes_dask/process_implementations/cubes/apply.py
LukeWeidenwalker: Added comments to document this!

Issue #22, opened 2022-12-20
`load_vector_cube` improvements


clausmichele:
I have some comments about the current `load_vector_cube` implementation:

1. If filename is actually a path to a file, this can't work https://github.com/Open-EO/openeo-processes-dask/blob/33bf285165b7346b2bd5cbe89db8aa92590c4316/openeo_processes_dask/process_implementations/cubes/experimental.py#L34 it should be instead:
```python
with open(filename,'r') as f:
    geometries = json.load(f)
```

2. Why don't we handle the case a geojson is provided within the graph? https://github.com/Open-EO/openeo-processes-dask/blob/33bf285165b7346b2bd5cbe89db8aa92590c4316/openeo_processes_dask/process_implementations/cubes/experimental.py#L51

3. My PR in geopandas was merged so this workaround could be removed: https://github.com/Open-EO/openeo-processes-dask/blob/33bf285165b7346b2bd5cbe89db8aa92590c4316/openeo_processes_dask/process_implementations/cubes/experimental.py#L55ù

Should I create a PR for improvements or is there a new implementation around already?

LukeWeidenwalker:
> 1. If `filename` is actually a path to a file, this can't work https://github.com/Open-EO/openeo-processes-dask/blob/33bf285165b7346b2bd5cbe89db8aa92590c4316/openeo_processes_dask/process_implementations/cubes/experimental.py#L34

Ah good spot with the filename, this codepath has clearly not been used in a while!

> Why don't we handle the case a geojson is provided within the graph?

Sorry, not sure what you mean - where in the graph would geojson additionally be provided?

> My PR in geopandas was merged so this workaround could be removed: 

Feel free to open a PR for 1 & 3! :)

clausmichele:
I was thinking at a similar process taking geometries like `aggregate_spatial`, where the geojson is specified within the graph. But anyway, this `load_vector_cube `process is not yet in the standard API yet, so this detail does not matter so much.
```json
    "aggregate1": {
      "process_id": "aggregate_spatial",
      "arguments": {
        "data": {
          "from_node": "reduce_bands"
        },
        "geometries": {
          "coordinates": [
            [
              [
                11.122454662230387,
                46.15311345671756
              ],
              [
                11.12356209638122,
                46.15135079436962
              ],
              [
                11.126409784197648,
                46.152026640492096
              ],
              [
                11.125460554925507,
                46.153789281194804
              ],
              [
                11.122454662230387,
[...]

PR #25, opened 2022-12-20
Load vector cube fix

clausmichele:
https://github.com/Open-EO/openeo-processes-dask/issues/22


clausmichele:
@LukeWeidenwalker apparently tests are failing since the `filename` was used to pass a json string to the process, how should we proceed? A json string should be the passed to the `geometries` parameter instead imho, but I'm not sure how this change would affect other stuff.

LukeWeidenwalker:
Ah this was done because the `read_vector` process node in the client injects them into the geometries, but we overlooked that `filename` can be an actual filename too 🤦 
Tbh, I think this `if not (URL or filename):` condition is unnecessary, it should be possible to directly inject the geometries. I'd suggest to then in `fit_regr_random_forest` to use that geometries argument and fix loading from filepath as you suggest!

Agree on removing the lines for geojson properties if they're no longer necessary!

Thanks!

Issue #27, opened 2022-12-21
`aggregate_temporal_period` missing some predefined periods


clausmichele:
The current implementation is missing some temporal periods which are not yet defined, since they are not pandas standard.
Additionally, if an openEO period is being selected and it's not present in the python code, there's no warning and it will raise an Exception due to the frequency variable missing.

Some good examples for Pandas periods:
https://regenerativetoday.com/a-complete-guide-to-time-series-analysis-in-pandas/

openEO process definition:
https://processes.openeo.org/#aggregate_temporal_period

Missing openEO periods:
```
    dekad: Ten day periods, counted per year with three periods per month (day 1 - 10, 11 - 20 and 21 - end of month). The third dekad of the month can range from 8 to 11 days. For example, the fourth dekad is Feb, 1 - Feb, 10 each year.
    tropical-season: Six month periods of the tropical seasons (November - April, May - October).
    decade: Ten year periods ([0-to-9 decade](https://en.wikipedia.org/wiki/Decade#0-to-9_decade)), from a year ending in a 0 to the next year ending in a 9.
    decade-ad: Ten year periods ([1-to-0 decade](https://en.wikipedia.org/wiki/Decade#1-to-0_decade)) better aligned with the anno Domini (AD) calendar era, from a year ending in a 1 to the next year ending in a 0.
```

an idea could be to just define custom periods based on the inputs and re use `aggregate_temporal`.

LukeWeidenwalker:
Agreed, this is an implementation gap from before that's never been addressed for lack of bandwidth - using pandas custom periods like you suggest sounds like a good idea to me! Would definitely be good to support all the predefined options.

ValentinaHutter:
https://github.com/Open-EO/openeo-processes-dask/pull/289

PR #28, opened 2022-12-21
Fix aggregate temporal

clausmichele:
Partial fix of https://github.com/Open-EO/openeo-processes-dask/issues/27

I added exceptions handling for `aggregate_temporal` and `aggregate_temporal_period`.

### DIFF
--- a/openeo_processes_dask/__init__.py
+++ b/openeo_processes_dask/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "2023.1.1"
+__version__ = "2023.1.2"
 
 import rioxarray as rio  # Required for the .rio accessor on xarrays.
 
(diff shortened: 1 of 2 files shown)


clausmichele:
Test is failing because the temporal dimension set in the test data is `t` but in the code we set it to `time` the default value. @LukeWeidenwalker let me know how we should address this.

LukeWeidenwalker:
Ah, this will be easily fixed after https://github.com/Open-EO/openeo-processes-dask/pull/20 is merged - I think just set it to `t` for the time being to satisfy the tests, we won't be doing any deployments of this until the accessor is merged anyways!

clausmichele:
Closing since @LukeWeidenwalker included the fixes in the commit https://github.com/Open-EO/openeo-processes-dask/commit/b3f1184c106b018021b09b0be195268c0960722b

PR #33, opened 2023-01-10
refactor: upgrade merge cubes to new callbacks

LukeWeidenwalker:
To be merged after #21 and #29 .
Updates merge_cubes to use the same callback structure and add tests for it.

I've tested dask scalability of this on a large-ish portion of boa_sentinel_2!


openeo_processes_dask/process_implementations/cubes/merge.py
```diff
@@ -10,126 +8,159 @@
...
+                len(overlap.only_in_cube1) == 0 and len(overlap.only_in_cube2) == 0
+                for overlap in overlap_per_shared_dim.values()
+            ]
+        )
+        if dims_have_no_label_diff:
+            # Example 3: All dimensions and their labels are equal
+            concat_both_cubes = (
+                xr.concat([cube1, cube2], dim=NEW_DIM_NAME)
+                .reindex({NEW_DIM_NAME: NEW_DIM_COORDS})
+                .chunk({NEW_DIM_NAME: -1, "x": "auto", "y": "auto"})
```
ValentinaHutter, line 61: Can the chunk of the NEW_DIM_NAME: -1, which means that this dimension is not chunked, become too large for dask? (Maybe this is handled in "x": "auto", "y": "auto" automatically, but I am not sure? Also, do we always expect "x" and "y" to be there?

ValentinaHutter (approved):
Really nice improvements!

openeo_processes_dask/process_implementations/cubes/merge.py
LukeWeidenwalker: Ah, no, NEW_DIM_NAME only ever has two coordinates, cube1 and cube2! Without this rechunking, the two cubes would live in different chunks, which doesn't work if we want to reduce over it!

Huh, actually good point on x and y. I included this, such that the config item of `array.chunk-size` is respected and chunks are broken down if too large - I'll have a look whether there's a way of doing this without specifying x and y explicitly!

LukeWeidenwalker: Updated!

PR #36, opened 2023-01-19
feat: Implement apply dimension

LukeWeidenwalker:
Closes https://eodc.atlassian.net/browse/EODCAPI-170

### DIFF
--- a/openeo_processes_dask/process_implementations/cubes/apply.py
+++ b/openeo_processes_dask/process_implementations/cubes/apply.py
@@ -1,5 +1,6 @@
 from typing import Callable, Optional
 
+import numpy as np
 import xarray as xr
 
 from openeo_processes_dask.exceptions import DimensionNotAvailable
@@ -47,7 +48,6 @@ def apply_dimension(
     positional_parameters = {"data": 0}
     named_parameters = {"context": context}
 
-    data.dims
     result = xr.apply_ufunc(
         process,
         data,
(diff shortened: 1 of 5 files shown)


tests/test_apply.py
```diff
@@ -36,3 +39,120 @@ def test_apply(temporal_interval, bounding_box, random_raster_data, process_regi
...
+    input_cube = create_fake_rastercube(
+        data=random_raster_data,
+        spatial_extent=bounding_box,
+        temporal_extent=temporal_interval,
+        bands=["B02", "B03", "B04", "B08"],
+        backend="dask",
+    )
+
+    _process = partial(
+        process_registry["add"], y=1, x=ParameterReference(from_parameter="data")
```
ValentinaHutter, line 58: The new `apply_dimension` looks really good and I think these tests are fine, but we should make sure that this also works with the `order` , `rearrange` and `sort` process as these cannot be used with `apply` or `reduce_dimension`.

openeo_processes_dask/process_implementations/cubes/apply.py
```diff
@@ -31,14 +32,43 @@ def apply_dimension(
...
     target_dimension: Optional[str] = None,
     context: Optional[dict] = None,
-    **kwargs,
 ) -> RasterCube:
+    if context is None:
+        context = {}
+
     if dimension not in data.dims:
         raise DimensionNotAvailable(
             f"Provided dimension ({dimension}) not found in data.dims: {data.dims}"
```
ValentinaHutter, line 41: In the documentation the DimensionNotAvailable Message is specified as: "A dimension with the specified name does not exist." Anyway, it makes sense to explicitly list the available dimensions here, so maybe we can just stick to this error message. :)

tests/test_apply.py
LukeWeidenwalker: Will look into that in the other PR!

openeo_processes_dask/process_implementations/cubes/apply.py
LukeWeidenwalker: Yeah, we currently deviate from the specified messages in a few places, where it makes sense to report more details, think that's fine tbh!

PR #50, opened 2023-01-27
feat: add comparison processes

ValentinaHutter:


openeo_processes_dask/process_implementations/comparison.py
```diff
@@ -20,142 +19,111 @@
...
-        y_type in [float, int]
-    ):  # both arrays only contain numbers
-        if type(delta) in [float, int]:
-            ar_eq = abs(x - y) <= delta
-        else:
-            ar_eq = x == y
+    if delta:
+        ar_eq = np.isclose(x, y, atol=delta)
+    elif not case_sensitive:
+        ar_eq = np.char.lower(x) == np.char.lower(y)
```
SerRichard, line 54: Maybe ensure the if and elif here are both  covered in the test?

openeo_processes_dask/process_implementations/comparison.py
```diff
@@ -20,142 +19,111 @@
...
+    exclude_max: Optional[bool] = False,
+    reduce: Optional[bool] = False,
+):
     if x is None or min is None or max is None:
         return None
-    if lt(max, min):
-        return False
     if exclude_max:
-        bet = da.logical_and(gte(x, min, reduce=reduce), lt(x, max, reduce=reduce))
+        bet = np.logical_and(gte(x, y=min, reduce=reduce), lt(x, y=max, reduce=reduce))
```
SerRichard, line 126: again, would be nice to ensure coverage of this conditional

SerRichard (approved):
lgtm, only comment would be to try and exhaust all conditional cases. Not so familiar with the processes at this point, but the test set up looks good!

clausmichele:
What's the status of this PR? Can I help to proceed with this?

LukeWeidenwalker:
Let's look into this after the SRR!

clausmichele:
Hi @ValentinaHutter and @LukeWeidenwalker, it would be important for me to have this functionality merged and released asap. Please let me know if this is not possible, so that I can plan the workshop of Monday in a different way.

openeo_processes_dask/process_implementations/comparison.py
```diff
@@ -0,0 +1,140 @@
...
+        return False
+    return np.isinf(x)
+
+
+def is_valid(x: ArrayLike):
+    finite = np.logical_not(is_infinite(x))
+    return np.logical_and(notnull(x), finite)
+
+
+def is_nodata(x: ArrayLike):
```
LukeWeidenwalker, line 41: As discussed in https://github.com/Open-EO/openeo-processes-dask/pull/101 I think we should not support `is_nodata` and `is_nan`, because these will not be conformant to the openeo spec!

tests/test_comparison.py
```diff
@@ -0,0 +1,258 @@
...
+        data=random_raster_data,
+        spatial_extent=bounding_box,
+        temporal_extent=temporal_interval,
+        bands=["B02", "B03", "B04", "B08"],
+        backend="dask",
+    )
+
+    _process = partial(
+        process_registry["eq"].implementation,
+        y=0.5,
```
LukeWeidenwalker, line 164: This test has potential to be flaky: If any of the randomly generated data values is 0.5, the test will fail. You could leverage the fact that the random data is always between -100 and 100, so you could make this 200

openeo_processes_dask/process_implementations/comparison.py
```diff
@@ -0,0 +1,140 @@
...
+    delta: Optional[float] = None,
+    case_sensitive: Optional[bool] = True,
+):
+    if not type(x) in [np.ndarray, da.core.Array] and not type(y) in [
+        np.ndarray,
+        da.core.Array,
+    ]:
+        if is_nodata(x) or is_nodata(y):
+            return np.nan
+    if x is False or y is False:
```
LukeWeidenwalker, line 61: I think this line isn't correct, eq(False, False) should be True!

openeo_processes_dask/process_implementations/comparison.py
```diff
@@ -0,0 +1,140 @@
...
+        return False
+    if delta:
+        ar_eq = np.isclose(x, y, atol=delta)
+    elif not case_sensitive:
+        ar_eq = np.char.lower(x) == np.char.lower(y)
```
LukeWeidenwalker, line 66: This will fail on any incoming arrays that aren't string arrays

openeo_processes_dask/process_implementations/comparison.py
```diff
@@ -0,0 +1,140 @@
...
+        np.ndarray,
+        da.core.Array,
+    ]:
+        if is_nodata(x) or is_nodata(y):
+            return np.nan
+    if x is False or y is False:
+        return False
+    if delta:
+        ar_eq = np.isclose(x, y, atol=delta)
+    elif not case_sensitive:
```
LukeWeidenwalker, line 65: why is this an elif? There needs to be a check on whether the array is of dtype string first, only then should case_sensitive be accepted

openeo_processes_dask/process_implementations/comparison.py
```diff
@@ -0,0 +1,140 @@
...
+):
+    if not type(x) in [np.ndarray, da.core.Array] and not type(y) in [
+        np.ndarray,
+        da.core.Array,
+    ]:
+        if is_nodata(x) or is_nodata(y):
+            return np.nan
+    if x is False or y is False:
+        return False
+    if delta:
```
LukeWeidenwalker, line 63: This should only be performed if numbers are compared

openeo_processes_dask/process_implementations/comparison.py
```diff
@@ -0,0 +1,140 @@
...
+
+def is_nodata(x: ArrayLike):
+    return np.logical_not(is_valid(x))
+
+
+def is_nan(x: ArrayLike):
+    return is_nodata(x)
+
+
+def eq(
```
LukeWeidenwalker, line 49: Datetimes are currently not considered here, so I think this wants to be included or an error raised!

openeo_processes_dask/process_implementations/comparison.py
```diff
@@ -0,0 +1,140 @@
...
+
+def neq(
+    x: ArrayLike,
+    y: ArrayLike,
+    delta: Optional[float] = None,
+    case_sensitive: Optional[bool] = True,
+):
+    eq_val = eq(x, y, delta=delta, case_sensitive=case_sensitive)
+    return np.where(
+        np.logical_and(is_valid(x), is_valid(y)), np.logical_not(eq_val), np.nan
```
LukeWeidenwalker, line 80: This applies for all the processes, but I think this shouldn't use `is_valid` (which also considers infinity to be invalid), but just the `notnull` function

LukeWeidenwalker:
Also, one more comment is that we totally should test this with `merge_cubes`, otherwise we don't know whether it actually works for the usecase that Michele is probably interested in!

clausmichele:
For the time being it would be already good to test a boolean mask generated with a comparison process with a multiplication like (pseudocode):

data = [[10,10],[10,5]]

mask = (data == 10)

masked_data = data * mask

masked_data = [[10,10],[10,0]]

openeo_processes_dask/process_implementations/comparison.py
ValentinaHutter: Hm, I think this came from the example of eq(x = 0, y = false) => false in https://processes.openeo.org/#eq. 
I think, I had assumed that if one incoming value is false, the output should be as well. But I probably interpreted it wrong, what do you think?

openeo_processes_dask/process_implementations/comparison.py
ValentinaHutter: So, the question is if we want to consider infinity as a valid or invalid value, right? I am not sure if we want to compare two infinite values with each other, e.g.  np.inf == np.inf -> True. Do we want this behaviour?

openeo_processes_dask/process_implementations/comparison.py
ValentinaHutter: I now added a check for it :)

openeo_processes_dask/process_implementations/comparison.py
ValentinaHutter: added a check, thank you for pointing it out!

openeo_processes_dask/process_implementations/comparison.py
ValentinaHutter: thanks for pointing it out, added an exception

openeo_processes_dask/process_implementations/comparison.py
ValentinaHutter: ah, you are right! removed them :)

ValentinaHutter:
> For the time being it would be already good to test a boolean mask generated with a comparison process with a multiplication like (pseudocode):
> 
> data = [[10,10],[10,5]]
> 
> mask = (data == 10)
> 
> masked_data = data * mask
> 
> masked_data = [[10,10],[10,0]]

Thank you, I will try to finish the updates by tomorrow. When does the workshow you mention take place? Is it next Monday (22nd)? :)

clausmichele:
Yes next Monday, but I would have to share the content a bit in advance with them.

tests/test_comparison.py
ValentinaHutter: good spot!

openeo_processes_dask/process_implementations/comparison.py
```diff
@@ -0,0 +1,127 @@
...
+    if is_number(x, y):
+        if not is_valid(x) or not is_valid(y):
+            return np.nan
+        if (
+            isinstance(x, bool)
+            and not isinstance(y, bool)
+            or isinstance(y, bool)
+            and not isinstance(x, bool)
+        ):
+            return False
```
ValentinaHutter, line 52: https://github.com/Open-EO/openeo-processes-dask/pull/50#discussion_r1193779887

I now added this check to make sure that the example of  eq(x = 0, y = false) => false works, but there is one problem left: 
I do this check for inputs that only consist of one value, e.g. int, float, bool. As soon as I want to do this for an array and check if it is a bool array, it might not work, because numpy easily converts a bool array into a float array. In that case, False would be set to 0 and the example would not work.

openeo_processes_dask/process_implementations/comparison.py
LukeWeidenwalker: > I had assumed that if one incoming value is false the output should be as well

I think you're thinking of the short circuiting that is required for the boolean processes (e.g. `and` being immediately False if the first value is False), but I don't think it's appropriate here!

The example `eq(x = 0, y = false) => false` is thus because the integer 0 is not equal to the boolean False

openeo_processes_dask/process_implementations/comparison.py
LukeWeidenwalker: Hmm, I think there's two questions here:
1) Is infinity a valid number in `is_valid`?
-> No, according to `is_valid` it isn't
2) is infinity null?
-> Also no, it's not nodata

I think the comparison processes just want to use `notnull` to check for nullness, rather than the `is_valid` process.

I'm not 100% sure on this, but since infinity is represented as a constant in IEEE 754 (see https://numpy.org/devdocs/reference/constants.html#numpy.Inf), I think `np.inf == np.inf -> True` seems right to me?

openeo_processes_dask/process_implementations/comparison.py
ValentinaHutter: @LukeWeidenwalker I tried implementing the eq(x = 0, y = false) => false example with an explicit check for the bool type if the input value only consists of one number - as soon as there are numpy or dask arrays, I would not check it explicitely, is this fine for you? or should I add a check for the numpy array dtype?

openeo_processes_dask/process_implementations/comparison.py
ValentinaHutter: okay, I now did some additional tests with numpy and inf comparison, and I think we can stick to how numpy handles this. I will push a new version asap :)

ValentinaHutter:
The current version shows one problem that now occured, as I tested `merge_cubes` with the `overlap_resolver` `eq`: The result of the `merge_cubes` process is of type `xr.DataArray` and loads the data into memory. I am not sure, why this happens within `merge_cubes`, as the `eq` process works as expected with `apply`. Need to investigate further.

LukeWeidenwalker:
Btw, I just realised, comparison of temporal stuff is deprecated with 2.0: https://github.com/Open-EO/PSC/issues/21, so we don't need to bother implementing this! :)

ValentinaHutter:
> Btw, I just realised, comparison of temporal stuff is deprecated with 2.0: [Open-EO/PSC#21](https://github.com/Open-EO/PSC/issues/21), so we don't need to bother implementing this! :)

Ah, thanks for pointing this out, then I will quickly update the check for the datetime again!

clausmichele:
Hi @ValentinaHutter and @LukeWeidenwalker, today it would be the last day to merge and make available the comparison processes and let me prepare the notebook for Monday

LukeWeidenwalker:
> Hi @ValentinaHutter and @LukeWeidenwalker, today it would be the last day to merge and make available the comparison processes and let me prepare the notebook for Monday

Morning - sorry, just saw your message, yesterday was a public holiday in Austria! @ValentinaHutter made more changes to this on Wednesday, which I'll review and iterate on again today. However, I don't think you really need us to merge&release these changes to use them. If you're happy with the current level of implementation, couldn't you just grab the code and stick it into your process registry manually?

clausmichele:
Unfortunately it's not a viable option, I would have to ask 20+ users to install a non released version from source and I wouldn't have the time for testing it now.

LukeWeidenwalker (approved):
@ValentinaHutter I've made a few more changes, and found another bug in merge_cubes! Just FYI, `not_null` wasn't dask-friendly yet, so now I was able to replace `da.where` with `np.where` again.

Think this looks good now, thanks for adding all the tests, these were really useful to have in place already! :)

@clausmichele: I'll merge this now and cut a release for you to use in your demo!

PR #65, opened 2023-02-21
feat: add array order processes

ValentinaHutter:


tests/test_arrays.py
```diff
@@ -201,6 +201,89 @@ def test_array_labels():
...
+        (
+            [6, -1, 2, np.nan, 7, 4, np.nan, 8, 3, 9, 9],
+            False,
+            False,
+            [3, 6, 9, 10, 7, 4, 0, 5, 8, 2, 1],
+        ),
+    ],
+)
+def test_order(data, asc, nodata, expected):
+    assert (order(data=data, asc=asc, nodata=nodata) == expected).all()
```
LukeWeidenwalker, line 234: This is a nitpick, but for these comparisons I like to use `np.testing.assert_array_equal`, because it does all kinds of useful checks and forces dask arrays to compute!

openeo_processes_dask/process_implementations/arrays.py
```diff
@@ -182,3 +182,75 @@ def array_labels(data: ArrayLike) -> ArrayLike:
...
+
+def order(
+    data: ArrayLike,
+    asc: Optional[bool] = True,
+    nodata: Optional[bool] = None,
+    axis: Optional[int] = None,
+):
+    if isinstance(data, list):
+        data = np.asarray(data)
+    if len(data) == 0:
```
LukeWeidenwalker, line 195: I think this should just return an empty array, not a nan-value

openeo_processes_dask/process_implementations/arrays.py
```diff
@@ -182,3 +182,75 @@ def array_labels(data: ArrayLike) -> ArrayLike:
...
+        return np.nan
+    if axis is None:
+        axis = 0
+    if asc:
+        permutation_idxs = np.argsort(data, kind="mergesort", axis=axis)
+    else:  # [::-1] not possible
+        permutation_idxs = np.argsort(
+            -data, kind="mergesort", axis=axis
```
LukeWeidenwalker, line 203: This only works with numerical types, but will fail if trying to sort string or datetime arrays!
Why not just use `np.flip` on the found indices?

ValentinaHutter: I had used flip before, but switched to this approach because of the following side effect: 
Using the example in the documentation, order(data = [6,-1,2,null,7,4,null,8,3,9,9], asc = false, nodata = false) should result in [3, 6, 9, 10, 7, 4, 0, 5, 8, 2, 1].
With np.flip our result becomes [6, 3, 10, 9, 7, 4, 0, 5, 8, 2, 1], so the values that are the same are in the opposite order. Switching these after flip was used does not seem too convenient to me, what do you think?

LukeWeidenwalker: Ah, thanks for this - hmm, what we could do would be to count the number of nans before, only flip the array[index_last_nan:] and stich together again? Then we should have strings/datetimes covered too!

ValentinaHutter: I think what you are referring to would result in an array like [3, 6, 10, 9, 7, 4, 0, 5, 8, 2, 1], so the nan values would be handled. If there are values in the data twice, (like the "9" in the example), the order of these is also flipped. (The result in the example should be [3, 6, 10, 9, 10, 7, 4, 0, 5, 8, 2, 1]

LukeWeidenwalker: Hmm. @m-mohr what do you think about relaxing the `Ties will be left in their original ordering` requirement in the process spec for `order`? Is this strictly necessary for some usecase? Seems like implementation on our end would be simpler if this didn't have to be guaranteed.

m-mohr: To get comparable and reproducible results, you can't let this undefined.

openeo_processes_dask/process_implementations/arrays.py
```diff
@@ -182,3 +182,104 @@ def array_labels(data: ArrayLike) -> ArrayLike:
...
+    source_transposed_axis: int = None,
+):
+    if len(data) == 0:
+        return data
+    if isinstance(data, list):
+        data = np.asarray(data)
+    if isinstance(order, list):
+        order = np.asarray(order)
+
+    if len(data.shape) != len(order.shape):
```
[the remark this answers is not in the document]
LukeWeidenwalker: Hmm - the spec seems ambiguous about that to me, since all the examples have an equal number of axes. We could add extra axes and broadcast the `order` array such that it fits the `data` array. I agree that it's a possible enhancement, but adding this would need more tests for higher-dimensional arrays. I think we can simplify our lives here and just throw an error saying that this isn't supported until someone actually needs exactly this?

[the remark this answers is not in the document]
m-mohr: Regardless of what type the elements are, the array is just re-arranged according to the order on the first level.

So if for example you have `rearrange(data = [[2,3], [4,3]], order=[1,0])`, you'd get `[[4,3], [2,3]]`. Hope that helps.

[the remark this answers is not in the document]
LukeWeidenwalker: > Regardless of what type the elements are, the array is just re-arranged according to the order on the first level.
> 
> So if for example you have `rearrange(data = [[2,3], [4,3]], order=[1,0])`, you'd get `[[4,3], [2,3]]`. Hope that helps.

Thanks for this comment - so order is always expected to only have a single axis?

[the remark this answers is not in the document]
m-mohr: Yes, indeed. It is of type `array<integer>` so can't be multi-dimensional.

openeo_processes_dask/process_implementations/arrays.py
LukeWeidenwalker: Okay, that's fair, so we cannot drop this requirement completely - what about changing it such that ties will be in the original ordering if `asc=True` and flipped otherwise?
Otherwise I currently fail to see a way to do this such that arrays of strings or datetimes can also be ordered, we'd have to throw an error in these cases.

m-mohr: @LukeWeidenwalker That's a good point, I think we can clarify that as I think that was the original intention anyway. Please open an issue in openeo-processes and I'll prepare a PR for 2.0.

LukeWeidenwalker: Opened https://github.com/Open-EO/openeo-processes/issues/409

LukeWeidenwalker:
@ValentinaHutter Update from https://github.com/Open-EO/openeo-processes/issues/409: ordering of ties is now implementation-dependent, so we can go ahead with the `np.flip` implementation! :)

PR #76, opened 2023-03-01
feat: add filter_temporal process

clausmichele:
Implemented `filter_temporal`. I will add tests soon.


openeo_processes_dask/process_implementations/cubes/_filter.py
```diff
@@ -15,9 +16,40 @@ def filter_bbox(data: RasterCube, extent: BoundingBox) -> RasterCube:
...
+                f"A dimension with the specified name: {dimension} does not exist."
+            )
+        applicable_temporal_dimension = dimension
+    else:
+        if not temporal_dims:
+            raise DimensionNotAvailable(
+                f"No temporal dimension detected on dataset. Available dimensions: {data.dims}"
+            )
+        if len(temporal_dims) > 1:
+            raise TooManyDimensions(
```
LukeWeidenwalker, line 35: This wants to be imported too

openeo_processes_dask/process_implementations/cubes/_filter.py
```diff
@@ -15,9 +16,40 @@ def filter_bbox(data: RasterCube, extent: BoundingBox) -> RasterCube:
...
 ) -> RasterCube:
-    raise NotImplementedError()
+    temporal_dims = data.openeo.temporal_dims
+
+    if dimension is not None:
+        if dimension not in data.dims:
+            raise DimensionNotAvailable(
+                f"A dimension with the specified name: {dimension} does not exist."
+            )
+        applicable_temporal_dimension = dimension
```
LukeWeidenwalker, line 28: Should we raise an exception here if `applicable_temporal_dimension` is not in `temporal_dims`? Or let the user override this, but at least log a warning?

openeo_processes_dask/process_implementations/cubes/_filter.py
```diff
@@ -15,9 +16,40 @@ def filter_bbox(data: RasterCube, extent: BoundingBox) -> RasterCube:
...
+            )
+        if len(temporal_dims) > 1:
+            raise TooManyDimensions(
+                f"The data cube contains multiple temporal dimensions: {temporal_dims}. The parameter `dimension` must be specified."
+            )
+    applicable_temporal_dimension = temporal_dims[0]
+
+    start_time = extent[0]
+    if start_time is not None:
+        start_time = start_time.to_numpy()
```
LukeWeidenwalker, line 42: This is only possible if starttime is either `DateTime` or `Year`, since `Date` and `Time` are ambiguous. I'm not entirely sure whether this wants to be done here. Would it make sense to immediately disambiguate temporal intervals in the parser when possible and just always throw an error otherwise? 

I.e. if `start` is `2023-03-02T15:34:17+00:00` and `end` is `16:34:17+00:00`, assume that the same date is meant and disambiguate `end` to `2023-03-02T16:34:17+00:00`.
Otherwise we'd possibly be duplicating this disambiguation logic across multiple temporal processes.
wdyt?

tests/test_filter.py
```diff
@@ -0,0 +1,41 @@
+from functools import partial
```
LukeWeidenwalker, line 1: np: unused import

LukeWeidenwalker (changes requested):
Thanks for adding the tests - looking good overall, just a few comments!

openeo_processes_dask/process_implementations/cubes/_filter.py
clausmichele: Indeed, if possible I would pass to dask-processes directly a numpy object, which would make it more portable and less dependant on the parser code.
For Date, I would just parse it like:
input: `2023-03-02` -> output: `2023-03-02T00:00:00+00:00` (check my PR in the parser, it's doing this)

Anyway, maybe we should firstly implement the simple scenarios (`DateTime`, where `Date` is also mapped to `DateTime`) and later on check if we can support `Time` only, which means supoprting `pendulum.Time` as an index variable in xArray.

openeo_processes_dask/process_implementations/cubes/_filter.py
LukeWeidenwalker: Looks good, thanks!

tests/test_filter.py
LukeWeidenwalker: Oh, sorry - I meant "np" as in "nitpick", not "numpy" :D 
The line that's unused is `from functools import partial`

clausmichele: Ok solved

openeo_processes_dask/process_implementations/cubes/_filter.py
clausmichele: @LukeWeidenwalker how do we proceed here? Raise an error if we try to use a temporal class which is not supported at the moment?

LukeWeidenwalker: > Raise an error if we try to use a temporal class which is not supported at the moment?

Yeah, let's do it like this, we can extend this in the future!

Btw I've just pulled in the changes from your PR in the parser, so can use them here too!

clausmichele: @LukeWeidenwalker I'll be on vacation until March 27. If you would like to close this PR before modify it as you prefer :)

LukeWeidenwalker: Ah, thanks for letting me know, I'll do that! - have a nice holiday 🌴

LukeWeidenwalker (approved):
Okay, finished this off - added proper handling of these temporal intervals in the parser, these are now resolved to an unambiguous numpy datetime - thanks again! :)

PR #80, opened 2023-03-03
chore: Add git submodule for process specs

LukeWeidenwalker:
- Enables importing the json specs like this: `from openeo_processes_dask.specs import reduce_dimension`
- Adds build variants for experimental and ml processes


ValentinaHutter:
LGTM!

Just one question from my side: Do we want to have a connection between specs and implementations? I think for the specs it does not really make sense, since we also added process specs for processes that are implemented elsewhere (e.g. io processes), but for the implementations it could be a requirement, that the spec is also available... This might be something for the backlog, as we do not really need it now, just something I was wondering about :)

LukeWeidenwalker:
Ah, yes, I was thinking that as a further step, we could add a test that checks whether the spec is available for all the processes!

SerRichard (approved):
lgtm! will try installing specs tomorrow in the api and see how it goes

PR #83, opened 2023-03-06
Remove `scale` process from math.py

ValentinaHutter:
I do not know where the `scale` process initially came from, but it does not belong to the official `openeo processes` and it is basically equivalent to the `multiply` process. What do you think about removing it from the processes? Is there any reason to keep it? :)


LukeWeidenwalker:
Good find, agree, let's get rid of it - since there's no spec it should be exposed anyways ;)

LukeWeidenwalker (approved):

PR #96, opened 2023-04-28
feat: add `add_dimension` process

clausmichele:
Implementation of `add_dimension process`, copy and paste implementation from `openeo-processes-python` with an added test


openeo_processes_dask/process_implementations/cubes/general.py
```diff
@@ -32,3 +34,33 @@ def dimension_labels(data: RasterCube, dimension: str) -> RasterCube:
...
+       The data cube with a newly added dimension. The new dimension has exactly one dimension label.
+       All other dimensions remain unchanged.
+    """
+    if name in data.dims:
+        raise Exception(
+            f"DimensionExists - A dimension with the specified name already exists. The existing dimensions are: {data.dims}"
+        )
+    data_e = data.assign_coords(placeholder=label)
+    data_e = data_e.expand_dims("placeholder")
+    data_e = data_e.rename({"placeholder": name})
```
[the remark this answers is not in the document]
clausmichele: Yes, I modified the code as suggested.

ValentinaHutter:
Thanks, LGTM!

tests/general_checks.py
```diff
@@ -12,6 +13,7 @@ def general_output_checks(
     expected_results=None,
     verify_crs: bool = False,
     verify_attrs: bool = False,
+    expected_dims: list = [],
```
LukeWeidenwalker, line 16:
```suggestion
    expected_dims: list = None,
```
See [here](https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument) for why this wants to be avoided!

tests/general_checks.py
```diff
@@ -37,6 +39,11 @@ def general_output_checks(
             output_data, expected_results, equal_nan=True, rtol=rtol
         )
 
+    if expected_dims:
```
LukeWeidenwalker, line 42:
```suggestion
    if expected_dims is not None:
```
Could want to compare against an empty list

openeo_processes_dask/process_implementations/cubes/general.py
```diff
@@ -32,3 +34,32 @@ def dimension_labels(data: RasterCube, dimension: str) -> RasterCube:
             f"Provided dimension ({dimension}) not found in data.dims: {data.dims}"
         )
     return data.coords[dimension]
+
+
+def add_dimension(
+    data: RasterCube, name: str, label: str, type: Optional[str] = "other"
```
LukeWeidenwalker, line 40: `type` is currently ignored. Ideally this would be worked into the `.openeo` accessor in `_xr_interop.py`, as an extra list for each type of dimension (e.g. "known_spatial_dims", etc.) that the name of this dimension is appended to and considered when `.openeo.spatial_dims` is called.

LukeWeidenwalker:
Couple of comments - the spec would also need to be added to https://github.com/eodcgmbh/openeo-processes and the submodule updated to that commit!

clausmichele:
> 

Ok commited suggested changes and opened PR for openeo-processes here https://github.com/eodcgmbh/openeo-processes/pull/7

LukeWeidenwalker (changes requested):
Thanks for the updates! Since https://github.com/eodcgmbh/openeo-processes/pull/7 was merged, the only things left to do are bumping the submodule and logging a warning that `type` is currently not supported!

openeo_processes_dask/process_implementations/cubes/general.py
clausmichele: Done, see last commit.

LukeWeidenwalker (commented):
Awesome, this looks good now, thanks!

I've bumped the submodule, see [in the README](https://github.com/Open-EO/openeo-processes-dask#specs) for how to do it!

LukeWeidenwalker (approved):

PR #101, opened 2023-05-03
docs: add documentation page for nodata handling

LukeWeidenwalker:
I've summarised the options and my recommendation for how to deal with nodata going forward in this repo, would be great to get all your eyes on this!

### DIFF
--- a/docs/decisions/handle-nodata-in-rastercubes.md
+++ b/docs/decisions/handle-nodata-in-rastercubes.md
@@ -15,6 +15,7 @@ we have to do this:
 
 **Example B: Workaround we have to use to account for null values are returned by logic processes**
 
+```python
 nan_x = np.isnan(x)
 nan_y = np.isnan(y)
 xy = np.logical_and(x, y)
@@ -23,7 +24,6 @@ xy = np.where(~nan_mask, xy, np.nan)
 nan_mask = np.logical_and(nan_y, xy)
 xy = np.where(~nan_mask, xy, np.nan)
 return xy
-
 Note that the output array of Example A has the dtype `bool_`, whereas the output array of Example B has dtype `float64`. This is because the missing data value `np.nan` is only defined for dtype `float64` and thus the entire array needs to be upcast to `float64`.
 
 See these numpy snippets to confirm this:
--- a/docs/decisions/handle-nodata-in-rastercubes.md
+++ b/docs/decisions/handle-nodata-in-rastercubes.md
@@ -1,0 +1,69 @@
+# How to handle nodata values in Rastercubes
+
+## Context
+The logic processes as defined by OpenEO can return True/False/null. In numpy there is no sentinel value for missing data in boolean/integer/char arrays (like there is with NaN for float).
+
+See [this blog](https://jakevdp.github.io/PythonDataScienceHandbook/03.04-missing-values.html) for a good overview to missing data handling in Python. Also see [NEP-26](https://numpy.org/neps/nep-0026-missing-data-summary.html) for the state of this issue in the python community.
+
+To implement `and` as specified, instead of this:
+
+**Example A: What you'd expect from an implementation for a logical `and` process**
+
+logical_and = np.logical_and(x, y)
+
+we have to do this:
+
+**Example B: Workaround we have to use to account for null values are returned by logic processes**
+
+nan_x = np.isnan(x)
+nan_y = np.isnan(y)
+xy = np.logical_and(x, y)
+nan_mask = np.logical_and(nan_x, xy)
+xy = np.where(~nan_mask, xy, np.nan)
+nan_mask = np.logical_and(nan_y, xy)
+xy = np.where(~nan_mask, xy, np.nan)
+return xy
+
+Note that the output array of Example A has the dtype `bool_`, whereas the output array of Example B has dtype `float64`. This is because the missing data value `np.nan` is only defined for dtype `float64` and thus the entire array needs to be upcast to `float64`.
+
+See these numpy snippets to confirm this:
+
+>>> np.array([True, False])
+array([ True, False], dtype='bool')  # boolean array
+>>> np.array([True, False]).itemsize
+1  # one byte per item
+>>> np.array([True, False, np.nan])
+array([ 1., 0., nan], dtype='float64')  # casts to float64
+>>> np.array([True, False, np.nan]).itemsize
+8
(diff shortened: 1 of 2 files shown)


docs/decisions/handle-nodata-in-rastercubes.md
```diff
@@ -0,0 +1,69 @@
...
+
+nan_x = np.isnan(x)
+nan_y = np.isnan(y)
+xy = np.logical_and(x, y)
+nan_mask = np.logical_and(nan_x, xy)
+xy = np.where(~nan_mask, xy, np.nan)
+nan_mask = np.logical_and(nan_y, xy)
+xy = np.where(~nan_mask, xy, np.nan)
+return xy
+
```
clausmichele, line 26:
```suggestion

```python
nan_x = np.isnan(x)
nan_y = np.isnan(y)
xy = np.logical_and(x, y)
nan_mask = np.logical_and(nan_x, xy)
xy = np.where(~nan_mask, xy, np.nan)
nan_mask = np.logical_and(nan_y, xy)
xy = np.where(~nan_mask, xy, np.nan)
return xy
```

```

File: docs/decisions/handle-nodata-in-rastercubes.md [hunk 2]
```diff
@@ -0,0 +1,69 @@
...
+## Context
+The logic processes as defined by OpenEO can return True/False/null. In numpy there is no sentinel value for missing data in boolean/integer/char arrays (like there is with NaN for float).
+
+See [this blog](https://jakevdp.github.io/PythonDataScienceHandbook/03.04-missing-values.html) for a good overview to missing data handling in Python. Also see [NEP-26](https://numpy.org/neps/nep-0026-missing-data-summary.html) for the state of this issue in the python community.
+
+To implement `and` as specified, instead of this:
+
+**Example A: What you'd expect from an implementation for a logical `and` process**
+
+logical_and = np.logical_and(x, y)
```
clausmichele, line 12:
```suggestion
```python
logical_and = np.logical_and(x, y)
```
```

File: docs/decisions/handle-nodata-in-rastercubes.md [hunk 3]
```diff
@@ -0,0 +1,69 @@
...
+See these numpy snippets to confirm this:
+
+>>> np.array([True, False])
+array([ True, False], dtype='bool')  # boolean array
+>>> np.array([True, False]).itemsize
+1  # one byte per item
+>>> np.array([True, False, np.nan])
+array([ 1., 0., nan], dtype='float64')  # casts to float64
+>>> np.array([True, False, np.nan]).itemsize
+8
```
clausmichele, line 38:
```suggestion
```python
>>> np.array([True, False])
array([ True, False], dtype='bool')  # boolean array
>>> np.array([True, False]).itemsize
1  # one byte per item
>>> np.array([True, False, np.nan])
array([ 1., 0., nan], dtype='float64')  # casts to float64
>>> np.array([True, False, np.nan]).itemsize
8
```
```

clausmichele (commented):
Left some suggestion to style the code appropriately. Other than this looks good!

clausmichele:
Hi @LukeWeidenwalker, I don't know why the suggestions didn't include the closing parts of the code blocks which I added, probably it's a bug of GitHub.
Please add them at the end of every code block, you can see the preview here https://github.com/Open-EO/openeo-processes-dask/blob/add-nodata-docs/docs/decisions/handle-nodata-in-rastercubes.md

LukeWeidenwalker:
@clausmichele @ValentinaHutter @SerRichard I'm flip-flopping on this issue again. I'm just looking at the specification of `is_nodata` (part of the legacy core profile) and if we go with the xarray way I proposed yesterday, there's no way that we can ever implement this process, because np.nan and nodata will effectively mean the same thing. I talked to a colleague about this over lunch and his take was that the only way of properly representing this is to use masked arrays. This can't be done as and when needed, any array we load from STAC into an xarray DataArray would need to be converted into a masked array immediately. Every process would need to work on underlying masked arrays.

I shied away from this in the beginning because, well, changing the default array container is kind of a big change. We'll have to make extra sure that the operations defined on this actually work - also with dask. Plus there's the performance implication of operations on masked arrays being slower overall (sometimes by quite a bit). And, it's not entirely clear how this would interfere with my plans for UC8 using sparse arrays.

But at this stage I really don't see a better way of implementing this requirement correctly. Please let me know what you think, or whether I'm missing something.

clausmichele:
As far as I know each xarray object could have a nodata value assigned to in the metadata, and np.nan is a common one. https://corteva.github.io/rioxarray/stable/getting_started/nodata_management.html
Additionally, openEO process definitions are not written in stone and are open to suggestions and edits if a developer thinks it's necessary. So, in my opinion I wouldn't worry so much if we can't be 100% compliant with the specs, especially if the effort that requires to do that is too high compared to other important functionalities for a normal user.

Probably you could open an issue on `openeo-processes` for the `is_nodata` process and ask to VITO and SentinelHub how they implemented it.

LukeWeidenwalker:
I think regardless of how VITO and Sinergise have "implemented" this (in quotes, because from VITO I've understood that nodata is natively supported for all datatypes in geotrellis, they don't need to think about), I still have no feasible way of doing this in xarray/dask land.

The only feasible way I see of supporting the concept of `null` in openeo is to follow xarray and treat np.nan as equal to nodata. This begs the question of what to do about the `is_nodata` and `is_nan` processes, especially with respect to the upcoming "core profile" of processes. If we go ahead with the current plan, then we will not be able to support these two processes, because there will be no distinction in our backend. 

@m-mohr Tagging you for advice here. Would it be an option to exclude these two processes from the core profile for the federation? We should be able to provide `is_valid` though. I don't expect that treating nodata and NaN the same internally would make a big difference in terms of results and reproducability.

m-mohr:
The core profile will be agreed on in SAP09 with participation of all infrastructures.

By the way, "null" in openEO was always meant to be a placeholder for a nodata value chosen by the infrastructure. This can be nan, null, NA, -1, 255, ... whatever is suitable. It should be exposed in metadata though. As it's controlled behavior with informed users, I don't see an issue with is_nan and is_null.

clausmichele (approved):

LukeWeidenwalker:
@m-mohr Thanks for this input! I assume by metadata you mean something like https://github.com/stac-extensions/raster on the results?

m-mohr:
Yes, for example (alternatives are the old file extension or the new classification extension)

ValentinaHutter (approved):

PR #103, opened 2023-05-05
feat: Implementation of resample_spatial to reproject values

SerRichard:
Dask implementation of resample spatial using the odc.algo reproject function. I've tried to handle some trickier parts (like infering the affine transformation) with some helper utils. Ready for review when all boxes are ticked!

- [X] Reprojection
- [x] Resampling
- [x] Unit tests


LukeWeidenwalker:
Just to leave it here too: Since https://github.com/opendatacube/odc-geo/pull/88 came out, it would probably be good to see whether this already does this!

SerRichard:
> Just to leave it here too: Since [opendatacube/odc-geo#88](https://github.com/opendatacube/odc-geo/pull/88) came out, it would probably be good to see whether this already does this!

Switched implementation to use the odc.reproject function, much less code to maintain now!

@ValentinaHutter One thing I'm not sure on. Odc reproject will automatically resample the raster to the provided projection. In the case a user provides no resolution to resample, and only wants a reproject, should we force the resolution of the input cube? Or just use accept resolution as none? Not sure what the preference would be.

SerRichard:
@ValentinaHutter One other concern, we've no way or reliably enforcing the convention of the user specifying the target resolution in the units of the output crs. Scao recommended this could be enforced with accepted rangers for different CRsystems. Example, if the user incorrectly specifies 0.0009 as their output resolution, and the crs is in metres, this will run for longer. Not sure how much longer on dask, so may not be a huge concern.

ValentinaHutter (commented):
LGTM!

ValentinaHutter:
> > Just to leave it here too: Since [opendatacube/odc-geo#88](https://github.com/opendatacube/odc-geo/pull/88) came out, it would probably be good to see whether this already does this!
> 
> Switched implementation to use the odc.reproject function, much less code to maintain now!
> 
> @ValentinaHutter One thing I'm not sure on. Odc reproject will automatically resample the raster to the provided projection. In the case a user provides no resolution to resample, and only wants a reproject, should we force the resolution of the input cube? Or just use accept resolution as none? Not sure what the preference would be.

In https://processes.openeo.org/#resample_spatial the default resolution=0 and says that it `Doesn't change the resolution by default (0).` So, I think this means, we need to keep the input resolution?

ValentinaHutter:
> @ValentinaHutter One other concern, we've no way or reliably enforcing the convention of the user specifying the target resolution in the units of the output crs. Scao recommended this could be enforced with accepted rangers for different CRsystems. Example, if the user incorrectly specifies 0.0009 as their output resolution, and the crs is in metres, this will run for longer. Not sure how much longer on dask, so may not be a huge concern.

Hm, it sounds quite complicated to check this with different ranges for every crs, what do you think? Also, the definition says, that the resolution is `Specified in the units of the target projection.` So, the user should be aware of this, and we can probably assume, that it is in the required unit. Does `data.odc.reproject` also assume, that the unit of the resolution matches the output crs?

openeo_processes_dask/process_implementations/cubes/resample.py
```diff
@@ -0,0 +1,73 @@
...
+    "q3",
+]
+
+
+def resample_spatial(
+    data: RasterCube,
+    projection: Union[str, int] = None,
+    resolution: Union[str, int] = None,
+    method: str = "near",
+    align: str = "upper-left",
```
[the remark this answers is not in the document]
SerRichard: Just logged a warning with the use of the function

tests/test_resample.py
```diff
@@ -0,0 +1,53 @@
...
+        input_cube=input_cube,
+        output_cube=output_cube,
+        verify_attrs=False,
+        verify_crs=False,
+    )
+
+    assert output_cube.odc.spatial_dims == ("y", "x")
+    assert output_cube.rio.crs == CRS.from_user_input(output_crs)
+
+    if output_crs != "4326":
```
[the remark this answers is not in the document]
SerRichard: added :)

openeo_processes_dask/process_implementations/cubes/resample.py
```diff
@@ -0,0 +1,73 @@
...
+    if "longitude" in reprojected.dims:
+        reprojected = reprojected.rename({"longitude": "x"})
+
+    if "latitude" in reprojected.dims:
+        reprojected = reprojected.rename({"latitude": "y"})
+
+    reprojected.attrs["crs"] = data.rio.crs
+
+    # Undo odc specific re-ordering.
+    reprojected = reprojected.transpose("bands", "t", "x", "y")
```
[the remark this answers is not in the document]
SerRichard: added :)

openeo_processes_dask/process_implementations/cubes/resample.py
```diff
@@ -0,0 +1,79 @@
...
+    "med",
+    "q1",
+    "q3",
+]
+
+
+def resample_spatial(
+    data: RasterCube,
+    projection: Union[str, int] = None,
+    resolution: Union[str, int] = None,
```
LukeWeidenwalker, line 31: According to the spec this wants to be 0 by default!

openeo_processes_dask/process_implementations/cubes/resample.py
```diff
@@ -0,0 +1,79 @@
...
+        method = "nearest"
+
+    elif method not in resample_methods_list:
+        raise Exception(
+            f'Selected resampling method "{method}" is not available! Please select one of '
+            f"[{', '.join(resample_methods_list)}]"
+        )
+
+    # Re-order, this is specifically done for odc reproject
+    data_cp = data.transpose("bands", "t", "y", "x")
```
LukeWeidenwalker, line 51: This will fail if the dimensions are named differently - think you can use the `.openeo` accessor to find the names dynamically instead!

openeo_processes_dask/process_implementations/cubes/resample.py
```diff
@@ -0,0 +1,79 @@
...
+
+    if not projection:
+        projection = data_cp.rio.crs
+
+    try:
+        projection = CRS(projection)
```
LukeWeidenwalker, line 57: Maybe initialise this with `CRS.from_user_input`

openeo_processes_dask/process_implementations/cubes/resample.py
```diff
@@ -0,0 +1,79 @@
...
+def resample_spatial(
+    data: RasterCube,
+    projection: Union[str, int] = None,
+    resolution: Union[str, int] = None,
+    method: str = "near",
+    align: str = "upper-left",
+):
+    """Resamples the spatial dimensions (x,y) of the data cube to a specified resolution and/or warps the data cube to the target projection. At least resolution or projection must be specified."""
+
+    if align == "upper-left":
```
LukeWeidenwalker, line 37: 1) This should especially appear if the user chose a different option than the default
2) If we don't support it, shouldn't we just remove it from the spec we expose?

openeo_processes_dask/process_implementations/cubes/resample.py
LukeWeidenwalker, line 31: Why can this be a string?

LukeWeidenwalker (changes requested):
Sorry for the delay on this review - nice use of odc.geo!
Couple of comments!

openeo_processes_dask/process_implementations/cubes/resample.py
clausmichele: The projection can be a string if provided as WKT or PROJ string. If it's an int it should be an EPSG code.
The resolution maybe should be int and float instead of string?

openeo_processes_dask/process_implementations/cubes/resample.py
SerRichard: This function just wraps this call, so not sure of the benefit?

openeo_processes_dask/process_implementations/cubes/resample.py
SerRichard: Updated!

SerRichard: yeah my mistake, the resolution should be an int!

openeo_processes_dask/process_implementations/cubes/resample.py
SerRichard: Yeah happy to remove it!

openeo_processes_dask/process_implementations/cubes/resample.py
SerRichard: eh changed ;)

LukeWeidenwalker:
Notes from our debugging session today:

- We observed the following to happen with a reprojection to EQUI7: 
image
 It should be possible to quick fix this by reprojecting the bounding box and cropping. This seems to be an issue solely for EQUI7, we've tried reprojecting to a few other projections and didn't encounter the same problem.
- The transpose does not work if the dimension names have changed. We don't think this transpose is really required, so can just get rid of this

SerRichard:
Have some time today, so will pick this back up and try to resolve the retrojection with equi7 issue we discovered last month + resolve test issues

SerRichard:
Using odc.geo 0.4.1, I was unable to reproduce what we saw last month one both fake and real (sen2like output) data. Fixed odc minimum version to 0.4.1, fixed tests and accessor import for odc, and reviewed output arrays with @LukeWeidenwalker!

openeo_processes_dask/process_implementations/cubes/resample.py
```diff
@@ -0,0 +1,77 @@
...
+    "min",
+    "med",
+    "q1",
+    "q3",
+]
+
+
+def resample_spatial(
+    data: RasterCube,
+    projection: Union[str, int] = None,
```
LukeWeidenwalker, line 30:
```suggestion
    projection: Optional[Union[str, int]] = None,
```

openeo_processes_dask/process_implementations/cubes/resample.py
```diff
@@ -0,0 +1,77 @@
...
+
+    # Re-order, this is specifically done for odc reproject
+    data_cp = data.transpose(
+        data.openeo.band_dims[0],
+        data.openeo.temporal_dims[0],
+        data.openeo.y_dim,
+        data.openeo.x_dim,
+    )
+
+    if not projection:
```
LukeWeidenwalker, line 54:
```suggestion
    if projection is None:
```

openeo_processes_dask/process_implementations/cubes/resample.py
```diff
@@ -0,0 +1,77 @@
...
+        projection = data_cp.rio.crs
+
+    try:
+        projection = CRS.from_user_input(projection)
+    except CRSError as e:
+        raise CRSError(f"{projection} Can not be parsed to CRS.")
+
+    if not resolution:
```
LukeWeidenwalker, line 62:
```suggestion
    if resolution == 0:
```

Just to be explicit!

openeo_processes_dask/process_implementations/cubes/resample.py
```diff
@@ -0,0 +1,77 @@
...
+        data.openeo.x_dim,
+    )
+
+    if not projection:
+        projection = data_cp.rio.crs
+
+    try:
+        projection = CRS.from_user_input(projection)
+    except CRSError as e:
+        raise CRSError(f"{projection} Can not be parsed to CRS.")
```
LukeWeidenwalker, line 60:
```suggestion
        raise CRSError(f"Provided projection string: '{projection}' can not be parsed to CRS.") from e
```

Would include the original traceback here

openeo_processes_dask/process_implementations/cubes/resample.py
```diff
@@ -0,0 +1,77 @@
...
+
+    if not resolution:
+        resolution = resolution_from_affine(data_cp.odc.geobox.affine).x
+
+    reprojected = data_cp.odc.reproject(
+        how=projection, resolution=resolution, resampling=method
+    )
+
+    if "longitude" in reprojected.dims:
```
LukeWeidenwalker, line 69: I think this renaming should only happen if the original dimension name was x/y too.

LukeWeidenwalker (changes requested):
Looking good! Couple of minor comments, but once these are resolved we should be able to finally merge this!

openeo_processes_dask/process_implementations/cubes/utils.py
```diff
@@ -1,8 +1,8 @@
 try:
     import dask
-
 except ImportError:
     dask = None
+import math
```
LukeWeidenwalker, line 5: This looks unintentional :D

LukeWeidenwalker:
Ahh, two more things:
- the spec allows the `align` parameter, we should at least accept that and log a warning that it doesn't do anything currently
- the spec needs to be updated in openeo-processes

SerRichard:
spec added in eodcgmbh/openeo-processes

clausmichele:
@SerRichard I have a question: after putting so much work into this process implementation, did you ever test it with an openEO process graph? Because currently it is not imported here https://github.com/Open-EO/openeo-processes-dask/blob/main/openeo_processes_dask/process_implementations/cubes/__init__.py and therefore can't be used.

clausmichele:
Additionally, not exposing a parameter like `align` just because it's not supported it's a bad idea in my opinion.
First of all since it's a parameter with a default value, which could be added by default added to the process graph by the clients (Python client adds it).

@soxofaan does the Python client always add to the process graph all the optional parameters with default values?

soxofaan:
> @soxofaan does the Python client always add to the process graph all the optional parameters with default values?

no we don't always automatically add the default value of unset parameters, it depends a bit on details of `DataCube` method.

In case of `resample_spatial` however, we do always add the `align` parameter with default value "upper-left"

m-mohr:
> Additionally, not exposing a parameter like align just because it's not supported it's a bad idea in my opinion.

Actually, it's the opposite. It's meant to be that if something is not supported, it's not exposed. Process definitions should reflect the implementation status.

> In case of resample_spatial however, we do always add the align parameter with default value "upper-left"

I think this should be changed. It should only be sent if not the default value and supported by the back-end.

clausmichele:
Thanks @m-mohr, so as you said, the Python Client needs to be adjusted.

soxofaan:
-> https://github.com/Open-EO/openeo-python-client/issues/500

m-mohr:
Yes, you are right. If one value is supported, the backend should change the schema accordingly so that it requires exactly that value. If a parameter is not supported at all, it should be removed.

That still would require some kind of change in the Python client, I think.

PR #104, opened 2023-05-05
feat: add is_valid

LukeWeidenwalker:


openeo_processes_dask/process_implementations/logic.py
```diff
@@ -0,0 +1,10 @@
+from typing import Any
+
+import numpy as np
+import pandas as pd
+from numpy.typing import ArrayLike
+
+
+def is_valid(x: Any) -> ArrayLike:
+    null_mask = pd.isnull(np.asarray(x))
```
ValentinaHutter, line 9: Can you check if this is dask compliant? I think np.asarray() loads the values into memory?

LukeWeidenwalker: Ah, good spot, forgot about that - updated to just use the xarray code directly, think that should do fine actually!

PR #109, opened 2023-05-11
fix: fix merge_cubes parameter passing

LukeWeidenwalker:


ValentinaHutter:
Lgtm! 
I was wondering if we could add an example test where the context from merge_cubes is passed to the overlap_resolver.... but I think we can also just keep it as it is. :)

PR #118, opened 2023-06-13
add filter_bbox

clausmichele:
Adding the `filter_bbox` process


openeo_processes_dask/process_implementations/cubes/_filter.py
```diff
@@ -88,3 +91,69 @@ def filter_bands(data: RasterCube, bands: list[str] = None) -> RasterCube:
...
+    return aoi
+
+
+def reproject_bbox(extent: BoundingBox, output_crs: str) -> BoundingBox:
+    if (
+        extent is not None
+        and extent.south is not None
+        and extent.west is not None
+        and extent.north is not None
+        and extent.east is not None
```
LukeWeidenwalker, line 125: I think you shouldn't need this ifclause, because the `BoundingBox` pydantic model would have thrown a ValidationError already!

openeo_processes_dask/process_implementations/cubes/_filter.py
```diff
@@ -88,3 +91,69 @@ def filter_bands(data: RasterCube, bands: list[str] = None) -> RasterCube:
...
+            y_slice = slice(trasnformed_extent.north, trasnformed_extent.south)
+        else:
+            y_slice = slice(trasnformed_extent.south, trasnformed_extent.north)
+    if len(data.x) > 1:
+        if data.x[0] > data.x[1]:
+            x_slice = slice(trasnformed_extent.east, trasnformed_extent.west)
+        else:
+            x_slice = slice(trasnformed_extent.west, trasnformed_extent.east)
+
+    aoi = data.loc[{"y": y_slice, "x": x_slice}]
```
LukeWeidenwalker, line 115: You'll want to use the `.openeo` accessor here to make this work for different dimension names!

openeo_processes_dask/process_implementations/cubes/_filter.py
```diff
@@ -88,3 +91,69 @@ def filter_bands(data: RasterCube, bands: list[str] = None) -> RasterCube:
...
+    return aoi
+
+
+def reproject_bbox(extent: BoundingBox, output_crs: str) -> BoundingBox:
```
LukeWeidenwalker, line 119:
```suggestion
def _reproject_bbox(extent: BoundingBox, output_crs: str) -> BoundingBox:
```
Since this is a private method

LukeWeidenwalker: Also I'd call this `target_crs` instead!

openeo_processes_dask/process_implementations/cubes/_filter.py
```diff
@@ -88,3 +91,69 @@ def filter_bands(data: RasterCube, bands: list[str] = None) -> RasterCube:
...
+        raise Exception(f"Empty or non-valid bounding box provided! {extent}")
+        return
+    if extent.crs is not None:
+        source_crs = extent.crs
+    else:
+        source_crs = "EPSG:4326"
+
+    transformer = Transformer.from_crs(source_crs, output_crs, always_xy=True)
+
+    x_t = []
```
LukeWeidenwalker, line 143: Just for clarity:
```suggestion
    x_reprojected = []
    y_reprojected = []
```

openeo_processes_dask/process_implementations/cubes/_filter.py
```diff
@@ -88,3 +91,69 @@ def filter_bands(data: RasterCube, bands: list[str] = None) -> RasterCube:
...
+
+def filter_bbox(data: RasterCube, extent: BoundingBox) -> RasterCube:
+    try:
+        input_crs = str(data.rio.crs)
+    except Exception as e:
+        raise Exception(f"Not possible to estimate the input data projection! {e}")
+    trasnformed_extent = reproject_bbox(extent, input_crs)
+
+    # Check if the coordinates are increasing or decreasing
+    if len(data.y) > 1:
```
LukeWeidenwalker, line 104: There should be an else statement to this, otherwise the slice might be undefined when it's called later

clausmichele:
@LukeWeidenwalker I updated the code using the .openeo accessor. However I noticed that calling data.openeo.spatial_dims throws an exception if the data has a single spatial dimension (you can get to this state using reduce_dimension(dimension='x',reducer='mean') for example). So, I updated the accessor to handle this scenario.

Additionally, there's also another edge case I'm not sure how to solve, when there's a single value on x_dim or y_dim: we need to check if this coordinate is actually inside the bbox or not. If not, should we raise an exception?

LukeWeidenwalker:
@clausmichele I just merged #125, please let me know if this doesn't resolve the issue that was blocking this PR!

clausmichele:
@LukeWeidenwalker now it should be fine and ready to be merged, thanks!
There's still an edge case which I didn't have time to decide how to handle, but the process is already functional like this.
I will try to fix that later on in another PR.

LukeWeidenwalker (approved):
Thanks for the changes, this lgtm now!

PR #125, opened 2023-06-22
Make dimension types static in `.openeo`

LukeWeidenwalker:
Keep track of the types of dimensions manually and guess only at datacube initialisation, instead of reguessing each time!

### DIFF
--- a/openeo_processes_dask/process_implementations/cubes/_xr_interop.py
+++ b/openeo_processes_dask/process_implementations/cubes/_xr_interop.py
@@ -47,16 +47,44 @@ class OpenEOExtensionDa:
                 found_dims.append(self._obj.dims[i])
         return found_dims
 
+    def _get_existing_dims_and_pop_missing(self, expected_dims):
+        existing_dims = []
+        for i, dim in enumerate(expected_dims):
+            if dim in self._obj.dims:
+                existing_dims.append(dim)
+            else:
+                expected_dims.pop(i)
+        return existing_dims
+
     @property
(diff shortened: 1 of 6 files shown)


LukeWeidenwalker:
@clausmichele this is my proposed fix for the problem you raised in #118. If you could please let me know if this solves the issue, that would be much appreciated - thanks!

clausmichele:
Ok, but if we merge this, we would have to modify also all the processes that modify the datacube dimensions like:
reduce_dimension, drop_dimension -> remove the reduced dim from the list

there's also the ndvi process that reduces the bands dimension AFAIK: https://github.com/Open-EO/openeo-processes-dask/blob/6563c1809fc4108581b78763a970b3bea1d5d935/openeo_processes_dask/process_implementations/math.py#L334

Additionally the band dimension is hardcoded there, it should be changed to `data.openeo._bands_dims`.

Not sure if  there are also other processes to consider.

LukeWeidenwalker:
> Ok, but if we merge this, we would have to modify also all the processes that modify the datacube dimensions like: reduce_dimension, drop_dimension -> remove the reduced dim from the list

Hmm,  `_get_existing_dims_and_pop_missing` deletes dimensions from the lists that the xarray object no longer has, so I think this should be fine! We'd only have to manually deal with this when a dimension is added, but not when it's removed. I can't think of any other place where that happens than `add_dimension`, can you?

> there's also the ndvi process that reduces the bands dimension AFAIK:

True, tbh, the entire NDVI process looks a bit fishy to me, I'll consider that separately!

PR #127, opened 2023-06-27
feat: add `load_stac`

clausmichele:
First version of `load_stac` using `stackstac` to load items generated by queries using `pystac-client`. Another possibility would be using `odc-stac`, but it currently has some limitations which I explained here: https://github.com/opendatacube/odc-stac/issues/54#issuecomment-1602704281
Anyway, we could even consider to support both, depending on the user requirements.

Currently supports only STAC Collections provided by catalogs with the `/query` endpoint for fitlering.

### DIFF
--- a/openeo_processes_dask/process_implementations/cubes/load.py
+++ b/openeo_processes_dask/process_implementations/cubes/load.py
@@ -1,4 +1,5 @@
 import datetime
+import json
 from collections.abc import Iterator
 from pathlib import PurePosixPath
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
@@ -16,6 +17,7 @@ from openeo_processes_dask.process_implementations.cubes._filter import (
     _reproject_bbox,
     filter_bands,
     filter_bbox,
+    filter_temporal,
 )
 from openeo_processes_dask.process_implementations.data_model import RasterCube
 from openeo_processes_dask.process_implementations.exceptions import (
@@ -42,8 +44,6 @@ def load_stac(
     properties: Optional[dict] = None,
     **kwargs,
 ) -> RasterCube:
-    modifier = kwargs["modifier"] if "modifier" in kwargs else None
-
     def validate_stac(url):
         stac = stac_validator.StacValidate(url)
         is_valid_stac = stac.run()
@@ -126,18 +126,38 @@ def load_stac(
                 query_params["query"] = properties
 
             items = catalog.search(**query_params).item_collection()
-            stack = stackstac.stack(items)
+            assets = None
+            if bands is not None:
+                assets = bands
+            stack = stackstac.stack(items, assets=assets)
             if spatial_extent is not None:
                 stack = filter_bbox(stack, spatial_extent)
-            if bands is not None:
-                stack = filter_bands(stack, bands)
-
             return stack
+
         else:
             # Load the whole collection wihout filters
             raise Exception(
                 f"No parameters for filtering provided. Loading the whole STAC Collection is not supported yet."
             )
+
+    elif asset_type == "ITEM":
+        stac_api = pystac_client.stac_api_io.StacApiIO()
+        stac_dict = json.loads(stac_api.read_text(url))
+        stac_item = stac_api.stac_object_from_dict(stac_dict)
+
+        assets = None
+        if bands is not None:
+            assets = bands
+
+        stack = stackstac.stack(stac_item, assets=assets)
+
+        if spatial_extent is not None:
+            stack = filter_bbox(stack, spatial_extent)
+
+        if temporal_extent is not None:
+            stack = filter_temporal(stack, temporal_extent)
+        return stack
+
     else:
         raise Exception(
             f"The provided URL is a STAC {asset_type}, which is not yet supported. Please provide a valid URL to a STAC Collection."
--- a/openeo_processes_dask/process_implementations/cubes/load.py
+++ b/openeo_processes_dask/process_implementations/cubes/load.py
@@ -57,6 +57,29 @@ def _validate_stac(url):
     return asset_type
 
 
+def _search_for_parent_catalog(url):
+    parsed_url = urlparse(url)
+    root_url = parsed_url.scheme + "://" + parsed_url.netloc
+    catalog_url = root_url
+    url_parts = PurePosixPath(unquote(parsed_url.path)).parts
+    collection_id = url_parts[-1]
+    for p in url_parts:
+        if p != "/":
+            catalog_url = catalog_url + "/" + p
+        try:
+            asset_type = _validate_stac(catalog_url)
+        except Exception as e:
+            logger.debug(e)
+            continue
+        if asset_type == "CATALOG":
+            break
+    if asset_type != "CATALOG":
+        raise Exception(
+            "It was not possible to find the root STAC Catalog starting from the provided Collection."
+        )
+    return catalog_url
+
+
 def load_stac(
     url: str,
     spatial_extent: Optional[BoundingBox] = None,
@@ -70,28 +93,9 @@ def load_stac(
         # If query parameters are passed, try to get the parent Catalog if possible/exists, to use the /search endpoint
         if spatial_extent or temporal_extent or bands or properties:
             # If query parameters are passed, try to get the parent Catalog if possible/exists, to use the /search endpoint
-            # url = "https://earth-search.aws.element84.com/v0/collections/sentinel-s2-l2a-cogs"
-            parsed_url = urlparse(url)
-            root_url = parsed_url.scheme + "://" + parsed_url.netloc
-            catalog_url = root_url
-            url_parts = PurePosixPath(unquote(parsed_url.path)).parts
-            collection_id = url_parts[-1]
-            for p in url_parts:
-                if p != "/":
-                    catalog_url = catalog_url + "/" + p
-                try:
-                    asset_type = _validate_stac(catalog_url)
-                except Exception as e:
-                    logger.debug(e)
-                    continue
-                if asset_type == "CATALOG":
-                    break
-            if asset_type != "CATALOG":
-                raise Exception(
-                    "It was not possible to find the root STAC Catalog starting from the provided Collection."
-                )
+            catalog_url = _search_for_parent_catalog(url)
 
-            ## Check if we are connecting to Microsoft Planetary Computer, where we need to sign the connection
+            # Check if we are connecting to Microsoft Planetary Computer, where we need to sign the connection
             modifier = pc.sign_inplace if "planetarycomputer" in catalog_url else None
 
             catalog = pystac_client.Client.open(catalog_url, modifier=modifier)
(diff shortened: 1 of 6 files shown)


clausmichele:
@soxofaan maybe you could also have a look at this PR, since it can be related to how we handle STAC in the python client too.

openeo_processes_dask/process_implementations/cubes/load.py
```diff
@@ -0,0 +1,164 @@
...
+from pathlib import PurePosixPath
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+from urllib.parse import unquote, urljoin, urlparse
+
+import planetary_computer as pc
+import pyproj
+import pystac_client
+import stackstac
+import xarray as xr
+from openeo_pg_parser_networkx.pg_schema import *
```
soxofaan, line 13: I think star imports like this should be avoided, especially from a dependency package. I have no idea what this line imports

openeo_processes_dask/process_implementations/cubes/load.py
```diff
@@ -0,0 +1,164 @@
...
+    url: str,
+    spatial_extent: BoundingBox = None,
+    temporal_extent: Optional[
+        list[Union[str, datetime.datetime, datetime.date]]
+    ] = None,
+    bands: Optional[list[str]] = None,
+    properties: Optional[dict] = None,
+    **kwargs,
+) -> RasterCube:
+    def validate_stac(url):
```
soxofaan, line 47: I don't think this function should be nested within `load_stac`

openeo_processes_dask/process_implementations/cubes/load.py
```diff
@@ -0,0 +1,164 @@
...
+        if len(stac.message) == 1:
+            try:
+                asset_type = stac.message[0]["asset_type"]
+            except:
+                raise Exception(f"stac-validator returned an error: {stac.message}")
+        else:
+            raise Exception(
+                f"stac-validator returned multiple items, not supported yet. {stac.message}"
+            )
+        return is_valid_stac, asset_type
```
soxofaan, line 63: why return `is_valid_stac`? it will always be true, because an exception is thrown above if it is false

openeo_processes_dask/process_implementations/cubes/load.py
```diff
@@ -0,0 +1,164 @@
...
+            parsed_url = urlparse(url)
+            root_url = parsed_url.scheme + "://" + parsed_url.netloc
+            catalog_url = root_url
+            url_parts = PurePosixPath(unquote(parsed_url.path)).parts
+            collection_id = url_parts[-1]
+            for p in url_parts:
+                if p != "/":
+                    catalog_url = catalog_url + "/" + p
+                try:
+                    is_valid_stac, asset_type = validate_stac(catalog_url)
```
soxofaan, line 84: This is pretty confusing: `asset_type` (and `is_valid_stac`) are overwritten here. is that intentional?

openeo_processes_dask/process_implementations/cubes/load.py
```diff
@@ -0,0 +1,164 @@
...
+                        spatial_extent_4326.east,
+                        spatial_extent_4326.north,
+                    ]
+                    query_params["bbox"] = bbox
+                except Exception as e:
+                    raise Exception(f"Unable to parse the provided spatial extent: {e}")
+
+            if temporal_extent is not None:
+                query_params["datetime"] = [
+                    str(temporal_extent[0].to_numpy()),
```
soxofaan, line 121: `temporal_extent[0]` can be a `str` according to the type annotation: what is this `to_numpy()` and how is that supposed to work?

soxofaan (commented):
this is a bit unknown territory for me, so just some superficial notes

openeo_processes_dask/process_implementations/cubes/load.py
clausmichele: Yes, it is intentional, since from a STAC Collection url I look for the root Catalog url, where I can perform a query and obtain the items to load. Maybe it's not the best approach, I will try to get in contact with the pystac_client developers to see if there's a better solution.

LukeWeidenwalker (changes requested):
Thanks a lot for this - bunch of comments! 

On the question of stackstac vs odc-stac: 
- I'm generally rather indifferent to which library we end up using, but would be good to be deliberate about the choice!
- You mention metadata not being parsed in https://github.com/opendatacube/odc-stac/issues/54#issuecomment-1602704281 - just for my understanding, which specific metadata are you referring to here?
- In anticipation of UC8, we should consider how we'll limit this query to certain geometries. `odc-stac` can do this with the `geopolygon` parameter [here](https://odc-stac.readthedocs.io/en/latest/_api/odc.stac.load.html#odc-stac-load), are you aware of any equivalent functionality in stackstac? 

I'll do some more testing with EODC collections specifically today, so there might be more feedback coming up!

openeo_processes_dask/process_implementations/cubes/load.py
```diff
@@ -0,0 +1,168 @@
...
+                    asset_type = _validate_stac(catalog_url)
+                except Exception as e:
+                    logger.debug(e)
+                    continue
+                if asset_type == "CATALOG":
+                    break
+            if asset_type != "CATALOG":
+                raise Exception(
+                    "It was not possible to find the root STAC Catalog starting from the provided Collection."
+                )
```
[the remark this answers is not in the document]
clausmichele: Done

openeo_processes_dask/process_implementations/cubes/load.py
```diff
@@ -0,0 +1,168 @@
...
+        if bands is not None:
+            assets = bands
+
+        stack = stackstac.stack(stac_item, assets=assets)
+
+        if spatial_extent is not None:
+            stack = filter_bbox(stack, spatial_extent)
+
+        if temporal_extent is not None:
+            stack = filter_temporal(stack, temporal_extent)
```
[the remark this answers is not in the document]
clausmichele: Done

clausmichele:
> @clausmichele Thanks a lot for this - bunch of comments!
> 
> On the question of stackstac vs odc-stac:
> 
>     * I'm generally rather indifferent to which library we end up using, but would be good to be deliberate about the choice!
> 
>     * You mention metadata not being parsed in [clarification on difference between this library and stackstac? opendatacube/odc-stac#54 (comment)](https://github.com/opendatacube/odc-stac/issues/54#issuecomment-1602704281) - just for my understanding, which specific metadata are you referring to here?
> 

You can test the difference using this sample script:
```python
import odc.stac
import stackstac
import pystac_client
import planetary_computer as pc
import stackstac

URL = "https://planetarycomputer.microsoft.com/api/stac/v1"
catalog = pystac_client.Client.open(URL,modifier=pc.sign_inplace)
spatial_extent = {"west": 11.259613, "east": 11.406212, "south": 46.461019, "north": 46.522237}
bbox = [spatial_extent["west"],spatial_extent["south"],spatial_extent["east"],spatial_extent["north"]]
items = catalog.search(
    bbox=bbox,
    collections=["landsat-8-c2-l2"],
    datetime=["2021-01-01T00:00:00.000Z", "2023-12-01T00:00:00.000Z"]
).get_all_items()
print(len(items))

print("+"*80)
print("OUTPUT OF ODC-STAC")
print("+"*80)
odc_data = odc.stac.load(
    items,chunks={})
print(odc_data.SR_B1)

print("+"*80)
print("OUTPUT OF STACKSTAC")
print("+"*80)

stackstac_data = stackstac.stack(items)
print(stackstac_data.loc[dict(band='SR_B1')])
```

Output:

```python
79
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
OUTPUT OF ODC-STAC
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
<xarray.DataArray 'SR_B1' (time: 79, y: 13132, x: 11832)>
dask.array<SR_B1, shape=(79, 13132, 11832), dtype=float32, chunksize=(1, 13132, 11832), chunktype=numpy.ndarray>
Coordinates:
  * y            (y) float64 5.374e+06 5.374e+06 5.374e+06 ... 4.98e+06 4.98e+06
  * x            (x) float64 4.98e+05 4.98e+05 4.98e+05 ... 8.529e+05 8.529e+05
    spatial_ref  int32 32632
  * time         (time) datetime64[ns] 2021-01-04T10:04:24.413826 ... 2022-03...
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
OUTPUT OF STACKSTAC
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
C:\Users\mclaus\Documents\GitHub\stackstac\stackstac\prepare.py:369: UserWarning: The argument 'infer_datetime_format' is deprecated and will be removed in a future version. A strict version of it is now t
he default, see https://pandas.pydata.org/pdeps/0004-consistent-to-datetime-parsing.html. You can safely remove this argument.
  times = pd.to_datetime(
<xarray.DataArray 'stackstac-4a7f4a6696aa69b99ddd17edd9f4fb28' (time: 79,
                                                                y: 13132,
                                                                x: 11832)>
dask.array<getitem, shape=(79, 13132, 11832), dtype=float64, chunksize=(1, 1024, 1024), chunktype=numpy.ndarray>
Coordinates: (12/27)
  * time                         (time) datetime64[ns] 2021-01-04T10:04:24.41...
    id                           (time) <U31 'LC08_L2SP_193027_20210104_02_T1...
    band                         <U13 'SR_B1'
  * x                            (x) float64 4.98e+05 4.98e+05 ... 8.529e+05
  * y                            (y) float64 5.374e+06 5.374e+06 ... 4.98e+06
    instruments                  object {'tirs', 'oli'}
    ...                           ...
    title                        <U46 'Coastal/Aerosol Band (B1)'
    gsd                          float64 30.0
    common_name                  object 'coastal'
    center_wavelength            object 0.44
    full_width_half_max          object 0.02
    epsg                         int32 32632
Attributes:
    spec:        RasterSpec(epsg=32632, bounds=(497970.0, 4980270.0, 852930.0...
    crs:         epsg:32632
    transform:   | 30.00, 0.00, 497970.00|\n| 0.00,-30.00, 5374230.00|\n| 0.0...
    resolution:  30.0
```

Anyway, we can also easily support both libraries! But at the moment I need to release this version based on stackstac and open a new PR later on showing how to use odc-stac as well.

>     * In anticipation of UC8, we should consider how we'll limit this query to certain geometries. `odc-stac` can do this with the `geopolygon` parameter [here](https://odc-stac.readthedocs.io/en/latest/_api/odc.stac.load.html#odc-stac-load), are you aware of any equivalent functionality in stackstac?
> 
> 
> I'll do some more testing with EODC collections specifically today, so there might be more feedback coming up!

From what I see in the odc-stac docs, it is just using the bbox of the polygon, so nothing particularly difficult to implement for stackstac as well.

LukeWeidenwalker:
Ah, gotcha, thanks for the example!
Changes look good to me know, but I just realized that we're still missing the process spec, would be great if that can be added, then I'm happy to merge this!

clausmichele:
@LukeWeidenwalker PR for adding the spec is here: https://github.com/eodcgmbh/openeo-processes/pull/9

LukeWeidenwalker (approved):

PR #128, opened 2023-06-29
chore: change logging structure

LukeWeidenwalker:
Follow-up to conversation in https://github.com/Open-EO/openeo-python-client/pull/434#discussion_r1221658135

@clausmichele what do you think about this structure instead?

### DIFF
--- a/openeo_processes_dask/process_implementations/__init__.py
+++ b/openeo_processes_dask/process_implementations/__init__.py
@@ -26,3 +26,6 @@ except ImportError as e:
 import rioxarray as rio  # Required for the .rio accessor on xarrays.
 
 import openeo_processes_dask.process_implementations.cubes._xr_interop
+
+log_format = "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
+logging.basicConfig(level=logging.INFO, format=log_format)
--- a/openeo_processes_dask/process_implementations/__init__.py
+++ b/openeo_processes_dask/process_implementations/__init__.py
@@ -2,7 +2,6 @@ import logging
 
 logger = logging.getLogger(__name__)
 
-
 from .arrays import *
 from .comparison import *
 from .cubes import *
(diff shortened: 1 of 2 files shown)


clausmichele:
@LukeWeidenwalker:
1. I would init the logger before the try imports, where we also use the logger.
2. It would be nice to allow the init of the log_format and of the level by the user, if they would like to specify a different format or set the level differently from INFO.

openeo_processes_dask/process_implementations/__init__.py
```diff
@@ -26,3 +26,6 @@
 import rioxarray as rio  # Required for the .rio accessor on xarrays.
 
 import openeo_processes_dask.process_implementations.cubes._xr_interop
+
+log_format = "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
+logging.basicConfig(level=logging.INFO, format=log_format)
```
soxofaan, line 31: Never call `logging.basicConfig` (or any other logging setup code) from the top-level of a module.
That's the responsibility of an application's entrypoint (something like `main()`).

Doing it from here can not be undone easily and which is frustrating for the users of you library

openeo_processes_dask/process_implementations/core.py
```diff
@@ -74,10 +74,11 @@ def wrapper(
             if arg not in inspect.signature(f).parameters:
                 resolved_kwargs.pop(arg, None)
 
-        pretty_args = {k: type(v) for k, v in resolved_kwargs.items()}
-        logger.warning(f"Running process {f.__name__}")
-        logger.warning(f"kwargs: {pretty_args}")
-        logger.warning("-" * 80)
+        pretty_args = {k: v.__repr__()[:80] for k, v in resolved_kwargs.items()}
```
soxofaan, line 77:
```suggestion
        pretty_args = {k: repr(v)[:80] for k, v in resolved_kwargs.items()}
```
would be a more canonical way of doing this

openeo_processes_dask/process_implementations/__init__.py
LukeWeidenwalker: Ah, very good point, thanks so much for pointing this out, this is super useful!!

PR #134, opened 2023-07-14
chore: add support for python 3.11

clausmichele:


clausmichele:
@LukeWeidenwalker tests passed with Python 3.11 👍

LukeWeidenwalker:
Ah awesome, didn't expect it to be that easy, but we'll take that :D 
Thanks!

PR #139, opened 2023-07-17
feat: add fit_curve and predict_curve

LukeWeidenwalker:
I've now taken a shot at implementing `fit_curve` too, and thought I'd just start from scratch to see what the problem really is. With just the xarray built-in function `curvefit`, I get throughput of 1000pixels/sec with a vanilla OpenEO dask cluster (6 workers w/ 4CPUs and 12GB RAM each). Dask has added constant memory rechunking a while ago with their [P2P rechunking scheme](https://discourse.pangeo.io/t/rechunking-large-data-at-constant-memory-in-dask-experimental/3266), so I'm somewhat hopeful that this will be okay for larger datacubes too. I haven't tried a truly humonguous dataset yet (the largest I tried was 1million pixels, 2x of what was used in [the SRR2 notebook](https://github.com/openEOPlatform/SRR2_notebooks/blob/main/UC6%20-%20Forest%20Dynamics.ipynb), which took 30 minutes), but I'm inclined to not worry about the rechunking anymore unless it comes up again.

@clausmichele do you think this level of performance allows you to make progress on this usecase?

There's some open questions about the interface:
- I don't think `fit_curve` should be run through `apply_dimension` or `reduce_dimension`, as was changed in https://github.com/Open-EO/openeo-processes/issues/425, because this makes everything awkward. I've therefore added the `dimension` parameter back in.
- I don't know how to support `ignore_nodata` right now
- The inference (`predict_curve`) is fine to run as a reducer though! I haven't had a chance to take a closer look though, so might need changes there too!

The parameter unpacking I've only kind of eyeballed so far, could use a closer look to confirm that what I'm doing there makes sense!

Note on performance: Most of the wall time is spent in a `vectorize__wrapper` step - afaict this is where `apply_ufunc` vectorizes the function before passing it to scipy's curve_fit, and this work isn't parallelising. I've already tried pre-compiling the function with numba before, but to no avail. Not entirely sure how we'd go about speeding this up further, but maybe the current throughput is already good enough to demonstrate the usecase?
image

### DIFF
--- a/openeo_processes_dask/process_implementations/ml/curve_fitting.py
+++ b/openeo_processes_dask/process_implementations/ml/curve_fitting.py
@@ -26,7 +26,7 @@ def fit_curve(data: RasterCube, parameters: list, function: Callable, dimension:
     # so we do this to generate names locally
     parameters = {f"param_{i}": v for i, v in enumerate(parameters)}
 
-    # The dimension along which to predict cannot be chunked!
+    # The dimension along which to fit the curves cannot be chunked!
     rechunked_data = data.chunk({dimension: -1})
 
     def wrapper(f):
--- a/openeo_processes_dask/process_implementations/ml/curve_fitting.py
+++ b/openeo_processes_dask/process_implementations/ml/curve_fitting.py
@@ -34,9 +34,17 @@ def fit_curve(data: RasterCube, parameters: list, function: Callable, dimension:
 
     # .curvefit returns some extra information that isn't required by the OpenEO process
     # so we simply drop these here.
-    fit_result = rechunked_data.curvefit(
-        dimension, wrapper(function), p0=parameters, param_names=list(parameters.keys())
-    ).drop_dims(["cov_i", "cov_j"])
+    fit_result = (
+        rechunked_data.curvefit(
+            dimension,
+            wrapper(function),
+            p0=parameters,
+            param_names=list(parameters.keys()),
+        )
+        .drop_dims(["cov_i", "cov_j"])
+        .to_array()
+        .squeeze()
+    )
     return fit_result
 
 
--- a/openeo_processes_dask/process_implementations/ml/curve_fitting.py
+++ b/openeo_processes_dask/process_implementations/ml/curve_fitting.py
@@ -2,6 +2,7 @@ from typing import Callable, Optional
 
 import numpy as np
 import xarray as xr
+from numpy.typing import ArrayLike
 
 from openeo_processes_dask.process_implementations.cubes import apply_dimension
 from openeo_processes_dask.process_implementations.data_model import RasterCube
@@ -52,31 +53,36 @@ def fit_curve(data: RasterCube, parameters: list, function: Callable, dimension:
 
 
 def predict_curve(
-    data: RasterCube,
-    function: Callable,
     parameters: RasterCube,
+    function: Callable,
     dimension: str,
-    labels: Optional[list] = None,
+    labels: Optional[ArrayLike] = None,
 ):
-    # data: (x, y, t, bands)
     # parameters: (x, y, bands, param)
-    labels = data.coords[dimension]
-    if np.issubdtype(labels.coords[dimension].dtype, np.datetime64):
+    if np.issubdtype(labels.dtype, np.datetime64):
         labels = labels.astype(int)
 
     def wrapper(f):
         def _wrap(*args, **kwargs):
             return f(
-                parameters=args[0],
-                x=labels,
+                *args,
+                positional_parameters={"parameters": 0},
+                named_parameters={"x": labels},
+                **kwargs,
             )
 
         return _wrap
 
-    xr.apply_ufunc(
+    return xr.apply_ufunc(
         wrapper(function),
         parameters,
+        vectorize=True,
         input_core_dims=[["param"]],
         output_core_dims=[[dimension]],
-        dask="allowed",
+        dask="parallelized",
+        output_dtypes=[np.float32],
+        dask_gufunc_kwargs={
+            "allow_rechunk": True,
+            "output_sizes": {dimension: len(labels)},
+        },
     )
--- a/openeo_processes_dask/process_implementations/ml/curve_fitting.py
+++ b/openeo_processes_dask/process_implementations/ml/curve_fitting.py
@@ -1,4 +1,7 @@
-from typing import Callable
+from typing import Callable, Optional
+
+import numpy as np
+import xarray as xr
 
 from openeo_processes_dask.process_implementations.cubes import apply_dimension
 from openeo_processes_dask.process_implementations.data_model import RasterCube
@@ -53,7 +56,27 @@ def predict_curve(
     function: Callable,
     parameters: RasterCube,
     dimension: str,
-    labels: list,
+    labels: Optional[list] = None,
 ):
-    # TODO: constrain prediction to nodata values
-    pass
+    # data: (x, y, t, bands)
+    # parameters: (x, y, bands, param)
+    labels = data.coords[dimension]
+    if np.issubdtype(labels.coords[dimension].dtype, np.datetime64):
+        labels = labels.astype(int)
+
+    def wrapper(f):
+        def _wrap(*args, **kwargs):
+            return f(
+                parameters=args[0],
+                x=labels,
+            )
+
+        return _wrap
+
+    xr.apply_ufunc(
+        wrapper(function),
+        parameters,
+        input_core_dims=[["param"]],
+        output_core_dims=[[dimension]],
+        dask="allowed",
+    )
--- a/openeo_processes_dask/process_implementations/ml/curve_fitting.py
+++ b/openeo_processes_dask/process_implementations/ml/curve_fitting.py
@@ -57,6 +57,8 @@ def fit_curve(data: RasterCube, parameters: list, function: Callable, dimension:
         .transpose(*expected_dims_after)
     )
 
+    fit_result.attrs = data.attrs
+
     return fit_result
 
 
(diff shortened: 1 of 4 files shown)


clausmichele:
@LukeWeidenwalker can you clarify how many time steps and how many pixels in the spatial dimension are present in your test set?

LukeWeidenwalker:
bands: 5 t: 160 y: 1151 x: 953

> @LukeWeidenwalker can you clarify how many time steps and how many pixels in the spatial dimension are present in your test set?

`bands: 5 t: 160 y: 1151 x: 953` (1.63 GiB)

clausmichele:
What is the chunk size? We would also need to check what happens if we increase the size in the spatial domain.

Anyway, to make usage of this functionality it should work also with a larger area, like a Sentinel-2 tile (~10000x10000).
Possibilities:
- Allocate more resources depending on the requested area? Need to check if throwing more workers reduces the required time.
- Splitting the jobs, similarly to what the Aggregator does for large scale processing?

LukeWeidenwalker:
> What is the chunk size? We would also need to check what happens if we increase the size in the spatial domain.
> 
> Anyway, to make usage of this functionality it should work also with a larger area, like a Sentinel-2 tile (~10000x10000). 

Chunk size after rechunking was `(1, 160, 1151, 953)`!
The time seems to scale linearly as far as I can see - doing half the number of pixels took ~50% of the time. Therefore we should expect a full sentinel-2 tile to take 50h with the current setup. 

> * Allocate more resources depending on the requested area? Need to check if throwing more workers reduces the required time.

I will try doubling the dask cluster size and report back on how that changes runtime - although I don't expect that to help, because as I said above, most of the time is spent in single cores working on this vectorize__wrapper call. 

> * Splitting the jobs, similarly to what the Aggregator does for large scale processing?

That would totally work, only subject to resource limitations on our backend!

clausmichele:
Where is set the re-chunking size? From the experiments I did, I remember that overall it was performing much better using small chunks like 128x128, 256x256, ...
It would be nice if you could test how much time it takes with the same settings but smaller chunks, currently they're pretty large.

LukeWeidenwalker:
I've now done a run with 12 workers instead of 6 - I get a minor speedup, from 31min to 24min, but it does indeed look like these extra resources aren't being fully utilised in this vectorisation step.

I've also tried different chunksizes (still with 12 workers):
- 256x256: took 30 minutes, so actually a bit slower than the larger chunk
- 128x128: took 20 minutes, so a bit faster
- 64x64: took 20 minutes, pretty much the same as 128x128

Interestingly, with smaller chunks I get pretty much full CPU utilisation across these 12 workers, as there's more `vectorize__wrapper` calls, but each of them is smaller, so all workers can work on them in parallel.
image

So extrapolating from these, I'd expect a sentinel-2 tile to take ~30h.
lmk what you think!

LukeWeidenwalker:
I've also just launched and killed pre-emptively a few experiments on a 10000x10000 spatial extent - my dask cluster did not run out of memory handling this array along the way, so although it would take a fair bit longer, I'm fairly confident that this computation would run through!

clausmichele:
Thanks Lukas for the tests. I actually don't have my old ones any more, but they were showing that with "small" chunks I was getting better performance.
When you have time, could you provide me the code to call fit_curve as you're doing using the python client?
If predict works as well I could start testing them any time soon.

LukeWeidenwalker:
> Thanks Lukas for the tests. I actually don't have my old ones any more, but they were showing that with "small" chunks I was getting better performance. When you have time, could you provide me the code to call fit_curve as you're doing using the python client? If predict works as well I could start testing them any time soon.

Ah, I'm not calling this with the python client, just locally with a remote dask cluster attached. Are you suggesting that we merge this, deploy it to prod as an experimental process and you start testing on that? Or should we add predict_curve too first and then deploy?

LukeWeidenwalker:
@clausmichele @ValentinaHutter I've now also implemented `predict_curve`, so this PR is ready for a thorough review! Since you're both more familiar with the actual usecase, it would be great if you could both have a look and tell me whether there's anything else missing here! :)

ValentinaHutter (commented):
LGTM! There are just some things, I was wondering about - see comments :)

openeo_processes_dask/process_implementations/ml/curve_fitting.py
```diff
@@ -0,0 +1,110 @@
...
+    )
+
+    return fit_result
+
+
+def predict_curve(
+    parameters: RasterCube,
+    function: Callable,
+    dimension: str,
+    labels: Optional[ArrayLike] = None,
```
ValentinaHutter, line 67:
``` labels = null
The labels to predict values for. If no labels are given, predicts values only for no-data (null) values in the data cube. ```

Something like `if labels is None: labels = data[dimension].values` might be a start :)

Comment (ValentinaHutter) [reply openeo_processes_dask/process_implementations/ml/curve_fitting.py:67]: After coordinating with @LukeWeidenwalker about this, we agreed that it makes more sense to remove the default value `None`, because there is no data set to get the null values from. This means that the definition needs to be updated as well.

File: openeo_processes_dask/process_implementations/ml/curve_fitting.py [hunk 2]
```diff
@@ -0,0 +1,110 @@
...
+):
+    labels_were_datetime = False
+    dims_before = list(parameters.dims)
+
+    if np.issubdtype(labels.dtype, np.datetime64):
```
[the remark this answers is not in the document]
LukeWeidenwalker: Good point - have added an attempt to cast to np.datetime in all situations!

openeo_processes_dask/process_implementations/ml/curve_fitting.py
```diff
@@ -0,0 +1,110 @@
...
+            p0=parameters,
+            param_names=list(parameters.keys()),
+        )
+        .drop_dims(["cov_i", "cov_j"])
+        .to_array()
+        .squeeze()
+        .transpose(*expected_dims_after)
+    )
+
+    return fit_result
```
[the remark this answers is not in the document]
LukeWeidenwalker: Good point, I can't think of any, but just losing them doesn't seem preferable either - added!

openeo_processes_dask/process_implementations/ml/curve_fitting.py
```diff
@@ -0,0 +1,110 @@
...
+            dimension,
+            wrapper(function),
+            p0=parameters,
+            param_names=list(parameters.keys()),
+        )
+        .drop_dims(["cov_i", "cov_j"])
+        .to_array()
+        .squeeze()
+        .transpose(*expected_dims_after)
+    )
```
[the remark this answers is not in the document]
LukeWeidenwalker: Not sure what you mean - the values arrive at the fitFunction as floats already!

LukeWeidenwalker:
@clausmichele would be great if you get some time to have a look over this PR this week - once this is merged, you should be able to start testing the usecase in short order!

openeo_processes_dask/process_implementations/ml/curve_fitting.py
```diff
@@ -0,0 +1,118 @@
...
+            f"Provided dimension ({dimension}) not found in data.dims: {data.dims}"
+        )
+
+    dims_before = list(data.dims)
+
+    # In the spec, parameters is a list, but xr.curvefit requires names for them,
+    # so we do this to generate names locally
+    parameters = {f"param_{i}": v for i, v in enumerate(parameters)}
+
+    # The dimension along which to predict cannot be chunked!
```
clausmichele, line 29: We are in the `fit_curve` definition, did you mean:

> The dimension along which to **fit** cannot be chunked!

openeo_processes_dask/process_implementations/ml/curve_fitting.py
```diff
@@ -0,0 +1,118 @@
...
+    rechunked_data = data.chunk({dimension: -1})
```
clausmichele, line 30: Shouldn't we do the same in `predict_curve` as well?

clausmichele (commented):
@LukeWeidenwalker I left two comments, other than that it seems fine.

openeo_processes_dask/process_implementations/ml/curve_fitting.py
LukeWeidenwalker: I don't think it matters for predict, because there each timestep can be inferenced for independently of the other steps!

clausmichele: @LukeWeidenwalker that's true. But I was wondering if it would be faster when predicting on a datacube with many timesteps?

openeo_processes_dask/process_implementations/ml/curve_fitting.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # .curvefit returns some extra information that isn't required by the OpenEO process
+    # so we simply drop these here.
+    fit_result = (
+        rechunked_data.curvefit(
+            dimension,
+            wrapper(function),
+            p0=parameters,
+            param_names=list(parameters.keys()),
+        )
```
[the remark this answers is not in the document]
LukeWeidenwalker: I think this wants to be done at the level of load_collection already, which it currently isn't - I'll create an issue to extend it there!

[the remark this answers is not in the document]
LukeWeidenwalker: Created https://eodc.atlassian.net/browse/SDO2023-310 - feel free to pick up if you have time ;)

openeo_processes_dask/process_implementations/ml/curve_fitting.py
LukeWeidenwalker: Hmm - not sure tbh, I haven't profiled `predict_curve` at all yet - I think I'll merge and deploy this now so we can start a training run at least, and revisit this if performance of inference turns out to be a problem!

openeo_processes_dask/process_implementations/ml/curve_fitting.py
LukeWeidenwalker: Raised https://github.com/Open-EO/openeo-processes/issues/453 to change this, and have already changed it in the spec we expose!

PR #140, opened 2023-07-19
fix: special case for filter_temporal

clausmichele:
I found a special case when `filter_temporal` fails with 
`KeyError: Timestamp('2022-05-10 00:00:00')`

It is probably due to some not conform elements in the temporal dimension.

```python
import openeo
local_conn = LocalConnection('')


spatial_extent = {"west": 11.1, "east": 11.5, "south": 46.1, "north": 46.5}
datacube = local_conn.load_stac(url=url,
                    spatial_extent=spatial_extent)
temporal_extent = ['2022-05-10T00:00:00','2022-06-30T00:00:00']
temporal_slice = datacube.filter_temporal(temporal_extent)
temporal_slice.execute()
```

This PR fixes that special case.

@LukeWeidenwalker could you try that code and confirm that with this PR it solves the issue for you as well? You would need to install the latest openeo-python-client from source, since the one with load_stac it's not released yet.

I actually tried to create a test to reproduce this special case but I didn't manage to reproduce it.

### DIFF
--- a/openeo_processes_dask/process_implementations/cubes/_filter.py
+++ b/openeo_processes_dask/process_implementations/cubes/_filter.py
@@ -68,7 +68,7 @@ def filter_temporal(
         ]
     except:
         try:
-            data = data.where(~np.isnat(data.time), drop=True)
+            data = data.where(~np.isnat(data[applicable_temporal_dimension]), drop=True)
             filtered = data.loc[
                 {applicable_temporal_dimension: slice(start_time, end_time)}
             ]
--- a/openeo_processes_dask/process_implementations/cubes/_filter.py
+++ b/openeo_processes_dask/process_implementations/cubes/_filter.py
@@ -62,18 +62,10 @@ def filter_temporal(
         # The specified instance in time is excluded from the interval.
         # See https://processes.openeo.org/#filter_temporal
 
-    try:
+        data = data.where(~np.isnat(data[applicable_temporal_dimension]), drop=True)
         filtered = data.loc[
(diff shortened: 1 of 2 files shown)


LukeWeidenwalker:
Thanks, just a heads-up that I won't have time to look into this until next week!

LukeWeidenwalker:
Hey @clausmichele - I've now had a look into this, here's what I found:
- The data loaded from the STAC catalogue you provided has a `NaT` value on one of the temporal indices. 
- That makes the index non-monotonic
- Which somehow (not sure why) messes with the slicing operation you do in `filter_temporal`

I think the fix for this would look a bit different, something like `data.where(~np.isnat(data.time), drop=True).coords["time"]` before attempting the `loc` operation should do the trick!

clausmichele:
> Hey @clausmichele - I've now had a look into this, here's what I found:
> 
>     * The data loaded from the STAC catalogue you provided has a `NaT` value on one of the temporal indices.
> 
>     * That makes the index non-monotonic
> 
>     * Which somehow (not sure why) messes with the slicing operation you do in `filter_temporal`
> 
> 
> I think the fix for this would look a bit different, something like `data.where(~np.isnat(data.time), drop=True).coords["time"]` before attempting the `loc` operation should do the trick!

I know about the NaT in the data and thanks for providing another easier solution, I will try it soon!

clausmichele:
@LukeWeidenwalker I used your suggestion and tested it, it works fine! Much simpler now.

LukeWeidenwalker:
Thanks for this fix too! :)

PR #147, opened 2023-07-28
fix: add more robust version of ndvi

### LINKED ISSUE #126: NDVI process bugs
There are multiple things to fix in the NDVI procesess:
1. The bands dimension is hardcoded, it should get it from the .openeo accessor instead
2. It calls multiple times data["bands"].values. It is not necessary, we just need to get the list once and use it multiple times.

https://github.com/Open-EO/openeo-processes-dask/blob/6563c1809fc4108581b78763a970b3bea1d5d935/openeo_processes_dask/process_implementations/math.py#L331

LukeWeidenwalker:
closes #126


LukeWeidenwalker:
@clausmichele could use a review from you on this if you have time! :)

LukeWeidenwalker:
Or @ValentinaHutter now that she's back! :)

openeo_processes_dask/process_implementations/cubes/indices.py
```diff
@@ -0,0 +1,56 @@
...
+    red_band_dim = "common_name" if red not in available_bands else band_dim
+
+    nir_band = data.sel({nir_band_dim: nir})
+    red_band = data.sel({red_band_dim: red})
+
+    nd = normalized_difference(nir_band, red_band)
+    if target_band is not None:
+        if target_band in data.coords:
+            raise BandExists("A band with the specified target name exists.")
+        nd = nd.assign_coords(bands=target_band).expand_dims(target_band)
```
ValentinaHutter, line 54: I think this only works if band_dim is actually called "bands". 

See https://processes.openeo.org/#ndvi under 
Return Value: 
target_band is a string: The data cube keeps the same dimensions. The dimension properties remain unchanged, but the number of dimension labels for the dimension of type bands increases by one. The additional label is named as specified in target_band.

ValentinaHutter, line 54:
```suggestion
        nd = nd.assign_coords(**{band_dim:target_band}).expand_dims(band_dim)
```

tests/test_indices.py
```diff
@@ -0,0 +1,66 @@
...
+    cube_with_nir_unresolvable.common_name.data = np.array(["blue", "red"])
+
+    with pytest.raises(NirBandAmbiguous):
+        ndvi(cube_with_nir_unresolvable)
+
+    cube_with_red_unresolvable = cube_with_resolvable_coords
+    cube_with_red_unresolvable.common_name.data = np.array(["nir", "yellow"])
+
+    with pytest.raises(RedBandAmbiguous):
+        ndvi(cube_with_nir_unresolvable)
```
ValentinaHutter, line 60:
```suggestion
        ndvi(cube_with_red_unresolvable)
```

ValentinaHutter:
LGTM! just some minor comments - maybe also add a test to check if the target_band label is set correctly in the band_dim.

openeo_processes_dask/process_implementations/cubes/indices.py
```diff
@@ -0,0 +1,56 @@
...
+            red not in available_bands
+            and "common_name" in data.xindexes._coord_name_id.keys()
+            and red not in data.coords["common_name"].data
+        ):
+            raise RedBandAmbiguous(
+                "The Red band can't be resolved, please specify the specific Red band name."
+            )
+
+    nir_band_dim = "common_name" if nir not in available_bands else band_dim
+    red_band_dim = "common_name" if red not in available_bands else band_dim
```
ValentinaHutter, line 45: Nice work using the common_name here! 👏

tests/test_indices.py
LukeWeidenwalker: Ahh, good spot!!

openeo_processes_dask/process_implementations/cubes/indices.py
LukeWeidenwalker: Ah, good catch, have updated this to work slightly differently! :)

PR #148, opened 2023-08-02
fix: handle incompatible coords in merge_cubes

clausmichele:
@LukeWeidenwalker, in load_stac, stackstac loads many additional metadata as data variables/coordinates. It is an advantage when we want to have a complete overview on the dataset but in this case creates an issue when trying to merge different datasets, An example is in this notebook I recently created:
https://github.com/clausmichele/openeo-python-client/blob/add_stac_notebook/examples/notebooks/Client_Side_Processing/client_side_processing_STAC_data_fusion.ipynb

This fix removes the unnecessary coordinates if the merge fails and tries to merge the datacubes again. I spent a lot of time trying to find a more elegant solution, but I couldn't.

### DIFF
--- a/openeo_processes_dask/process_implementations/cubes/merge.py
+++ b/openeo_processes_dask/process_implementations/cubes/merge.py
@@ -113,6 +113,8 @@ def merge_cubes(
                     cube1 = cube1.to_dataset(cube1.openeo.band_dims[0])
                     cube2 = cube2.to_dataset(cube2.openeo.band_dims[0])
 
+                # compat="override" to deal with potentially conflicting coords
+                # see https://github.com/Open-EO/openeo-processes-dask/pull/148 for context
                 merged_cube = xr.combine_by_coords(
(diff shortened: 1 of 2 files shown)


LukeWeidenwalker:
Ah, thanks for reporting this - I had a similar problem in #147, but there I could just use `xr.merge`, which doesn't seem to have this problem. I'm still a bit confused on why this wouldn't work though, I would have imagined `combine_by_coords` to just create the missing coord - do you think this is something worth raising with xarray?

Regardless, happy to go for a workaround like you propose - if you could please also add a test case to cover this failure mode, that would be much appreciated - thanks!

clausmichele:
I tried also to write the two datasets to netCDFs, read them again with xArray and then it worked fine, which is so strange.
I agree that it would be good to create a minimal working example and open an issue for xarray.

LukeWeidenwalker:
Ah, good find - thanks for the fix!

PR #151, opened 2023-08-16
Fix linear scale range

clausmichele:
Adding the clip step before scaling.


LukeWeidenwalker:
Ah, thanks for this fix!!

PR #152, opened 2023-08-16
fix: support axis keyword in array_contains

### LINKED ISSUE #63: Support `axis` keyword for array_contains
`array_contains` is supposed to be a reducer, but doesn't currently support the required axis keyword.

LukeWeidenwalker:
Closes #63

### DIFF
--- a/tests/test_arrays.py
+++ b/tests/test_arrays.py
@@ -161,7 +161,12 @@ def test_array_contains(data, value, expected):
 
 def test_array_contains_axis():
     data = np.array([[4, 5, 6], [5, 7, 9]])
-    result = array_contains(data, 5, axis=1)
+
+    result_0 = array_contains(data, 5, axis=0)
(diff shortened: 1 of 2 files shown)


tests/test_arrays.py
```diff
@@ -159,6 +159,11 @@ def test_array_contains(data, value, expected):
     assert dask_result == expected or dask_result.compute() == expected
 
 
+def test_array_contains_axis():
+    data = np.array([[4, 5, 6], [5, 7, 9]])
+    result = array_contains(data, 5, axis=1)
+
```
ValentinaHutter, line 165: This test does not have an `assert`, yet.

ValentinaHutter (commented):
LGTM, but I think, since `array_contains` is labeled as a `reducer`, it might be useful to add a small test to run `reduce_dimension(reducer=array_contains)`.

LukeWeidenwalker:
Ah, good points - fixed both, thanks! :)

PR #155, opened 2023-08-23
Draft for labelled arrays

clausmichele:
First draft to support labelled arrays.

I modified only the processes required to run the following simple example. We would have to go through all the math processes and others as well if we want to implement this.

```python
from openeo.processes import pi

bands = ["red", "nir"]
    url=url,
    bands=bands,
    properties=properties,
)

from openeo.processes import array_element
def labeled_ndvi(x):
    b4 = array_element(x,label="red")
    b8 = array_element(x,label="nir")
    ndvi = (b8-b4)/(b8+b4)
    return ndvi

s2_ndvi = s2_datacube.reduce_dimension(dimension="band",reducer=labeled_ndvi)

s2_ndvi = s2_ndvi.execute()
```

### DIFF
--- a/openeo_processes_dask/process_implementations/arrays.py
+++ b/openeo_processes_dask/process_implementations/arrays.py
@@ -7,6 +7,7 @@ import numpy as np
 import pandas as pd
 import xarray as xr
 from numpy.typing import ArrayLike
+from openeo_pg_parser_networkx.pg_schema import DateTime
 from xarray.core.duck_array_ops import isnull, notnull
 
 from openeo_processes_dask.process_implementations.cubes.utils import _is_dask_array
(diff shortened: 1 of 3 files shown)


LukeWeidenwalker (changes requested):
Had a first look now - thanks for looking into this!
The general approach of just passing labels through to child processes looks fine to me. I'd recommend adding another reserved parameter to the list in `https://github.com/Open-EO/openeo-processes-dask/blob/main/openeo_processes_dask/process_implementations/core.py#L71C1-L72C1`. This removes the parameter if the subprocess can't do anything with it and we shouldn't need to touch all the math processes. Wdyt?

clausmichele:
@LukeWeidenwalker I followed your suggestion, it seems to work fine.
I also added a simple test based on array_element + label

clausmichele:
@ValentinaHutter @GeraldIr it would be good if you could review this and merge it!

GeraldIr (approved):
LGTM

clausmichele:
@ValentinaHutter @GeraldIr please have a look again at this PR so that it can be merged.

PR #164, opened 2023-09-29
fix: reduce_spatial

clausmichele:
`reduce_spatial` was not working due to wrong parameters. I fixed it and added a test.

### DIFF
--- a/openeo_processes_dask/process_implementations/cubes/reduce.py
+++ b/openeo_processes_dask/process_implementations/cubes/reduce.py
@@ -54,9 +54,8 @@ def reduce_spatial(
     spatial_dims = data.openeo.spatial_dims if data.openeo.spatial_dims else None
     return data.reduce(
         reducer,
-        dimension=spatial_dims,
+        dim=spatial_dims,
         keep_attrs=True,
-        context=context,
         positional_parameters=positional_parameters,
         named_parameters=named_parameters,
     )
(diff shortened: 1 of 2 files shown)


openeo_processes_dask/process_implementations/cubes/reduce.py
```diff
@@ -54,9 +54,8 @@ def reduce_spatial(
     spatial_dims = data.openeo.spatial_dims if data.openeo.spatial_dims else None
     return data.reduce(
         reducer,
-        dimension=spatial_dims,
+        dim=spatial_dims,
         keep_attrs=True,
-        context=context,
```
GeraldIr, line 59: Why is context being dropped here? To my knowledge this is passed downstream to the reducer function which might be using it no?

GeraldIr (commented):
1 comment, otherwise lgtm.

openeo_processes_dask/process_implementations/cubes/reduce.py
clausmichele: That's true, it was a mistake, fixed now.

clausmichele:
@GeraldIr thanks, I fixed what you commented.

openeo_processes_dask/process_implementations/cubes/reduce.py
ValentinaHutter: I think you can remove the context, as it gets passed to the subprocess in the `named_parameters`, then the tests should pass again :)

openeo_processes_dask/specs/openeo-processes
ValentinaHutter, line 1: Where are the changes to the submodule comming from? As they are not related to this process implementation, I do not think we need them in this PR, but maybe I missed something...

PR #166, opened 2023-09-29
fix: hardcoded band dim name

clausmichele:


GeraldIr (approved):
If this is tested, lgtm.

PR #170, opened 2023-10-09
Filter spatial

masawdah:
Hi team, 
together with @clausmichele have implemented spatial_filter process and added a test for it.


clausmichele:
This PR includes implementations of filter_spatial and mask_polygon. We tested them on full S2 tiles, you also may need to test them on the EODC back-end and report if you see some issues.

Current implementation supports only geoJSON vector-cubes, since currently the `load_vector_cube` process has not been implemted and we didn't discuss on which data structure to use for it (probably dask-geopandas?)

clausmichele:
@ValentinaHutter @GeraldIr can you start reviewing this PR?

openeo_processes_dask/process_implementations/cubes/_filter.py
GeraldIr, line 1: It seems that the dimensions defined at the beginning of the filter_spatial function are unused. I'm unsure if they need to be used or if it's leftover code, so i thought I'd mention it as these values are used in the other functions.

GeraldIr (commented):
Just the one question about unused variables, but that doesn't seem to impact the logic of the function, otherwise LGTM!

openeo_processes_dask/process_implementations/cubes/_filter.py
masawdah: Indeed they're leftover code. I'll remove them.

clausmichele:
@GeraldIr @ValentinaHutter I think it would be good to test the implementation at EODC with your cluster and with your data, to check how it behaves with large datasets.

PR #186, opened 2023-10-31
Apply kernel

VincentVerelst:
Added a first implementation of the apply_kernel process along with some tests.


clausmichele:
Hi @VincentVerelst, I've tried the code with a practical example and it seems fine.

```python

bands = ["red", "nir"]
    url=url,
    bands=bands,
    properties=properties,
)

b04 = s2_datacube.band("red")
b08 = s2_datacube.band("nir")
ndvi = (b08 - b04) / (b08 + b04)
ndvi_median = ndvi.reduce_dimension(dimension="time", reducer="median")
result_ndvi = ndvi_median.execute()
result_ndvi.plot.imshow(vmin=0,vmax=1,cmap="Greens")
```

Untitled

```python
import numpy as np
ndvi_median_blur = ndvi_median.apply_kernel(kernel=np.ones((5,5)),factor=1/25)
ndvi_median_blur_xr = ndvi_median_blur.execute()
ndvi_median_blur_xr.plot.imshow(vmin=0,vmax=1,cmap="Greens")
```

Untitled-1

soxofaan:
FYI @LukeWeidenwalker   we want to assign reviewers for this PR, but there is no option to do that in the right column:
image

clausmichele:
I do see a list of people, not sure if I have different rights on this repo:
image

m-mohr:
The spec says: "Each level of the kernel must have an uneven number of elements, otherwise the process throws a KernelDimensionsUneven exception.". If you don't throw an error the process spec for this implementation should explain what happens in this case.

GeraldIr:
@VincentVerelst Can you confirm what happens in the case of an uneven kernel and if it is in line with the spec, if yes I will merge it, otherwise I would request a kernel dimension check and throwing the correct exception.

VincentVerelst:
@GeraldIr , I still need to add the kernel dimension check, will keep you posted!

VincentVerelst:
Added the KernelDimensionsUneven exception.

Issue #124, opened 2023-06-19
Roadmap towards UC8


LukeWeidenwalker:
@clausmichele @aljacob I wanted to follow up on our call last Friday with a roadmap of concrete steps we'll need to finally close UC8.

Last year, because of the way the architecture was set up and there was no local execution environment, we could only test the whole workflow within the entire running system, so it was difficult to collaborate on this outside of EODC. With the rewrites to the parser and the processes, this should now be much easier, because we can essentially prototype the entire workflow without needing to have it exposed in a fully-fledged backend. As discussed on Friday, I'm summing up here what's still missing, so that @clausmichele  can start supporting us again!

Subissues:
- Support the `geometries` parameter in `load_collection` (or `load_stac`) by loading sparse arrays (https://github.com/pydata/sparse) (#120, #121)
	-  The problem with this usecase was always that we had no way to only load the data that we'd crop to later in `aggregate_spatial`. I'm attaching my proof of concept notebook for how to get around this by leveraging pydata/sparse arrays (super crude - @clausmichele please skim this quickly and let me know if you can work with this already, or if I need to tidy it up and add some more comments!). This has to be done at the point where data is loaded initially, because otherwise any operations between load_collection and aggregate_spatial will run over an extent much larger than necessary. 
	- @clausmichele : If you prototype this in `load_stac`, the rest of the pipeline should be unlocked to be worked on, even if this isn't live at EODC yet!
- Rewrite `aggregate_spatial` to work with sparse arrays and produce vector cubes with https://github.com/xarray-contrib/xvec (#119)
	- When we worked on this last year, we used geopandas+dask_geopandas to represent vector cubes somewhat haphazardly. This is actually not a good idea in the long term, because we'd have to handle raster and vector cubes separately everywhere (think `if isinstance(data, gpd.GeoDataFrame):` statements all over the place) - which would be annoying and make this unnecessarily hard to maintain! 
	- Xvec has since come out, which I find quite promising, because it should allow us to support vector cubes within the xarray paradigm. It has the added benefit that the implementation is directly motivated by Edzer's concept of the vector cube, so we should be quite closely aligned conceptually already. It should also be easier to keep processes dask-friendly that way, rather than relying on the extra dependency of geopandas/dask-geopandas.
	- Disclaimer: I haven't tried xvec yet, so all of this (especially how it behaves with dask) still needs to be confirmed!
- adapt `fit_regr_random_forest` to work on xvec vector cubes (#122)
[...]

PR #194, opened 2023-11-14
Aggregate spatial

masawdah:
Hi team, 
We have completed the implementation of the aggregate_spatial process. You can use the code snippet below to test it. The current output format is an x-vector cube as an xarray dataset, but it can also be configured as a DatArray, depending on the requirements of processes utilizing the output, such as fit_regr_random_forest.

```import numpy as np
import xarray as xr
import pandas as pd
from shapely.geometry import MultiPolygon, Polygon, mapping
import geopandas as gpd
import json


json_file = {
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "coordinates": [
          [
            [
              11.143181900944768,
              46.682468440575775
            ],
            [
              11.143181900944768,
              46.67892170969685
            ],
            [
              11.14948639229209,
              46.67892170969685
            ],
            [
              11.14948639229209,
              46.682468440575775
            ],
            [
              11.143181900944768,
              46.682468440575775
            ]
          ]
        ],
        "type": "Polygon"
      }
    },
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "coordinates": [
          [
            [
              11.181891477817032,
              46.69812317718774
            ],
            [
              11.176847884739885,
              46.697171913851975
            ],
            [
              11.173695639065556,
              46.69284777838911
            ],
            [
              11.178234872836669,
              46.69250183259447
            ],
            [
              11.181891477817032,
              46.69812317718774
            ]
          ]
        ],
        "type": "Polygon"
      }
    }
  ]
}

bbox =  [10.93311286, 46.66029087, 11.41500808, 46.85360212]


url = "https://planetarycomputer.microsoft.com/api/stac/v1/collections/sentinel-2-l2a"
spatial_extent =  {"west": bbox[0], "south": bbox[1], "east": bbox[2], "north": bbox[3], 'crs':'EPSG:4326'}

bands = ["B04","B03","B02","B08"] # B04 = red, B03 = green, B02 = blue, B08 = near infrared
properties = {"eo:cloud_cover": dict(lt=30)}


datacube = local_conn.load_stac(
    url=url,
    bands=bands,
    properties=properties,
)

# Geometries - Polygon
datacube = datacube.filter_spatial(json_file)

datacube = datacube.aggregate_spatial(json_file, reducer='mean')
datacube
```

Here if the output is xr.Dataset 

image
 

And this is as DataArray

image


clausmichele:
@masawdah in the example you mention a `json_file` but it's not defined.
Anyway, it would be good to provide a simple geometry in the code so that it's self contained and it can be executed without further inputs.

masawdah:
> @masawdah in the example you mention a `json_file` but it's not defined. Anyway, it would be good to provide a simple geometry in the code so that it's self contained and it can be executed without further inputs.

updated with geometries and bbox.

ValentinaHutter:
LGTM, thank you for working on this!

openeo_processes_dask/process_implementations/cubes/aggregate.py
```diff
@@ -109,11 +123,146 @@ def aggregate_temporal_period(
...
-    geometries: VectorCube,
+    geometries,
     reducer: Callable,
-    target_dimension: str = "result",
-    **kwargs,
+    chunk_size: int = 2,
 ) -> VectorCube:
-    raise NotImplementedError
+    t_dim = data.openeo.temporal_dims
+    t_dim_name = t_dim[0]
```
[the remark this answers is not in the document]
ValentinaHutter: It would make sense to also add 

t_dim_name = None if len(t_dim) == 0 else t_dim[0] 

otherwise this line fails when the temporal dimension is not available. This would also need to be considered in Line 231 and 243.

ValentinaHutter:
If you want to run the pre commit manually, you could `pip install pre-commit` and then try `pre-commit run --all-files`. :)

openeo_processes_dask/process_implementations/cubes/aggregate.py
```diff
@@ -109,11 +123,132 @@ def aggregate_temporal_period(
...
+
+    times = list(data[t_dim_name].values)
+
+    data_vars = {}
+    for key in keys_items.keys():
+        data_vars[key] = (["geometry", t_dim], df[keys_items[key]])
+
+    ## Create VectorCube
+    vec_cube = xr.Dataset(
+        data_vars=data_vars, coords=dict(geometry=df.geometry, time=times)
```
ValentinaHutter, line 251: Instead of naming the temporal dimension "time" per default, it makes sense to name it after the temporal dimension in the input data, what do you think? 
I think, this could be achieved by something like

`coords= {"geometry": df.geometry, t_dim: times}`

masawdah: Indeed, it makes sense to keep the name as it is. Unfortunately, I currently don't have access to my VM due to internal maintenance. However, you can go ahead and implement the new modifications. 

The name of time dimension has already been extracted in the beginning with `t_dim_name = t_dim[0]`, so we can reuse it here.

ValentinaHutter: Ah, I was looking into it, but you were faster with your updates, I now also added a test for a dataset without time.

ValentinaHutter:
Thanks for working on this! I will now release a new openeo-processes-dask version and continue testing on larger scales!

openeo_processes_dask/process_implementations/cubes/aggregate.py
```diff
@@ -109,11 +124,140 @@ def aggregate_temporal_period(
...
+    mask = rasterio.features.geometry_mask(
+        geometry_array, out_shape=(y_dim_size, x_dim_size), transform=transform
+    )
+
+    if t_dim is not None:
+        mask = np.expand_dims(mask, axis=data_dims.index(t_dim))
+    if b_dim is not None:
+        mask = np.expand_dims(mask, axis=data_dims.index(b_dim))
+
+    masked_data = data * mask
```
martinfleis, line 159: Hey, I got here from https://github.com/xarray-contrib/xvec/pull/52 where @masawdah opened a PR with a variation of this. 

The same code as is here was there as well but I believe it is wrong. `mask` is a boolean array with False where the `geometry_array` has been burned in, the rest is True. Line 159 results in an array where all the values are preserved except those where the geometry is, where the values are now 0 (not missing!).

Any reduction of this masked array it therefore likely wrong. 

Even if you passed `invert=True` to `rasterio.features.geometry_mask` to get True indicating the location of geometry and False the area that should be taken away, the line 159 still keeps 0 where the geometry is not, therefore mean, median etc will be wrong. You would need to create a mask resulting in missing values here I believe.

I may be wrong as this is clearly reviewed PR but I don't think that is the case. The tests here also do not verify any correctness of values and this code indeed produces correctly formed cube. Just with completely wrong results.

I have refactored the logic in that `xvec` PR and verified the result against other implementation so you can check how it is done there. Or we may be able to get the built-in functionality of xvec to the state where you can use it directly, if you dare to depend on version 0.2.0 of it :).

clausmichele:
@masawdah @ValentinaHutter, I also noticed something that needs to be adjusted.
You can run the first example adding `datacube.execute()` at the end and inspect the result.
1. The result is an xarray.Dataset, but we agreed to use DataArray as a common data format.
2. The result is not Dask based, somewhere the load/computation of the data is being triggered. We need to avoid that and do it only when explicitly requested.

openeo_processes_dask/process_implementations/cubes/aggregate.py
masawdah: @martinfleis  Thanks for pointing this out. Actually, this work was done simultaneously with [xarray-contrib/xvec#52](https://github.com/xarray-contrib/xvec/pull/52) . The intention behind it was to replace this effort with a built-in functionality of xvec once it is ready.  

I saw your refactoring for the implementation, and it looks great. Let's double-check once again to confirm the best :)

masawdah:
> @masawdah @ValentinaHutter, I also noticed something that needs to be adjusted. You can run the first example adding `datacube.execute()` at the end and inspect the result.
> 
> 1. The result is an xarray.Dataset, but we agreed to use DataArray as a common data format.
> 2. The result is not Dask based, somewhere the load/computation of the data is being triggered. We need to avoid that and do it only when explicitly requested.

I'm not sure if I missed something about the output format but it mentioned here https://github.com/Open-EO/openeo-processes-dask/issues/124#issuecomment-1810331949 xarray.dataset. but it would be easy to switch.

clausmichele:
Please be aware that now the conversion from xarray.Dataset to xarray.DataArray is using the `.to_dataarray()` method, the `.to_array()` is deprecated. See here: https://docs.xarray.dev/en/stable/generated/xarray.Dataset.to_dataarray.html#xarray.Dataset.to_dataarray

image

masawdah:
@clausmichele  @ValentinaHutter  I have tested the new implementation of the built-in functionality of xvec for spatial aggregation, and it performed well. It successfully handled the points mentioned by Michele.

We can replace the existing functionality with this new implementation upon its release. 

Thanks @martinfleis  

  
image

clausmichele:
@masawdah there's a new release of xvec including your contribution https://github.com/xarray-contrib/xvec/releases/tag/v0.2.0
You can create a new PR now

PR #203, opened 2023-12-13
[load_stac] allow load from static Collection

clausmichele:
Allow to use `load_stac` with a static STAC Collection. It will be necessary for the test suite, since we plan to host the STAC Collection on GitHub (no API).

You can test it with the one I prepared:

```python

url = "https://raw.githubusercontent.com/clausmichele/openeo-test-suite/eurac_process_graphs/src/openeo_test_suite/data/SENTINEL2_L2A_SAMPLE/SENTINEL2_L2A_SAMPLE.json"
bands = ["B04"]

s2_cube = local_conn.load_stac(url=url,
   bands=bands,
)
s2_cube.execute()
```

### DIFF
--- a/openeo_processes_dask/process_implementations/cubes/load.py
+++ b/openeo_processes_dask/process_implementations/cubes/load.py
@@ -80,11 +80,12 @@ def _search_for_parent_catalog(url):
         )
     return catalog_url, collection_id
 
+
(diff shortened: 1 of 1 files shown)


clausmichele:
I need to clarify if the self-contained STAC Collection is actually valid, I will re open the PR once clarified.

PR #205, opened 2023-12-14
array_find in reverse order gives wrong indices

m-mohr:
array_find in reverse order gives wrong indices.

Doesn't quite work for integer no-data values (99999) yet...


ValentinaHutter:
just realized I resolved this in a separate PR now...

PR #206, opened 2023-12-14
Clip correctly in linear_scale_range when inputMax < inputMin

m-mohr:
Clip correctly in linear_scale_range when inputMax < inputMin

Found by the new test suite and want to make you aware. I won't have time to add tests, this is more a bug report rather than a full PR.


clausmichele:
@m-mohr where is this special behaviour defined? https://processes.openeo.org/#clip

PR #210, opened 2023-12-14
Add array_append

m-mohr:
Not exactly sure how to add array_append to the submodule...


SerRichard:
I've added array append to this submodule, think we'll put some time aside in the next week and close this and a few other PRs

PR #220, opened 2024-02-02
Bug in mask polygon 

SerRichard:
Culprit line in mask polygon is indexing the y_dim and x_dim values, they're strings so this will fail if you the dimensions are named latitude or longitude. :smile:


ValentinaHutter:
lgtm!

PR #221, opened 2024-02-02
bump submodule to include new process specs

SerRichard:
Bumping submodule to include new specs for array_append and also adaptive_thresholding


ValentinaHutter:
lgtm! :)

Issue #228, opened 2024-02-21
hardcoded spatial dims labels in `resample_spatial`


clausmichele:
Instead of `"x"` and `"y"` we should use `data.openeo.x_dim` and `data.openeo.y_dim`.
I will fix it in a PR.

https://github.com/Open-EO/openeo-processes-dask/blob/main/openeo_processes_dask/process_implementations/cubes/resample.py#L78-L82

ValentinaHutter:
merged - thank you!

PR #233, opened 2024-02-28
Fix resample spatial

clausmichele:
Fix for https://github.com/Open-EO/openeo-processes-dask/issues/228

Note: I add to add the default python language in the pre-commit configuration otherwise it didn't work in my environment.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,7 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
+default_language_version:
+    python: python3
 repos:
 -   repo: https://github.com/asottile/pyupgrade
     rev: v3.10.1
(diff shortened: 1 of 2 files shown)


ValentinaHutter:
Thanks for checking this in the process implementation and bringing it up! 

Anyhow, I have been wondering - does it make sense to rename the spatial coordinates after a reprojection in general? 
If data is reprojected from UTM to WGS84, the coordinates would be renamed to latitude and longitude by the reprojection and could stay like that, couldn't they? 

What do you think about it?

clausmichele:
The dimension labels should not be changed by this process, as specified in the process definition: https://processes.openeo.org/#resample_spatial

ValentinaHutter:
Oh, right - thanks for pointing it out - it just feels a bit counter-intuitive, doesn't it? But this is then more a question on the process definition level :)

Issue #243, opened 2024-04-03
Module openeo_processes_dask.specs has no attribute 'load_collection'


jzvolensky:
How to reproduce:

1. Clone the main branch with
```git clone git@github.com:Open-EO/openeo-processes-dask.git```

2. Follow the install instructions:
```poetry install --all-extras```

3. Add OpenEO Local Processing
```poetry add "openeo[localprocessing]"```

4. Try importing and running a script e.g.
```from openeo.local import LocalConnection```

5. Script fails with:
```AttributeError: module 'openeo_processes_dask.specs' has no attribute 'load_collection'```

It is necessary to clone with `--recurse-submodules`:
```git clone --recurse-submodules git@github.com:Open-EO/openeo-processes-dask.git```

This is currently not mentioned in the installation instructions and needs to be added. I can add it to the readme together with another fix currently in progress. 

Thanks!

ValentinaHutter:
Thanks for pointing this out! 

In order to run openeo-processes-dask implementations or unit tests, you do not necessarily need the git submodule to be available - therefore, we only mentioned it at the bottom of the https://github.com/Open-EO/openeo-processes-dask?tab=readme-ov-file#specs if you want to make use of it in connection with the local processing, this needs to be available, so, I think it would also make sense to add a section on this in the local processing docs, what do you think, @clausmichele ?

clausmichele:
Hi @ValentinaHutter, in my opinion the submodule should always be cloned, since the best way to develop new processes is using real process graphs, which facilitates a lot the testing of various scenarios when the process is chained with other ones.

@jzvolensky included a couple of lines in his PR here: https://github.com/Open-EO/openeo-processes-dask/pull/244/files

ValentinaHutter:
Sure, we can include it in the openeo-processes-dask docs, looks fine! 
But wouldn't it also make sense to include this in the client side processing documentation? https://open-eo.github.io/openeo-python-client/cookbook/localprocessing.html#installation

clausmichele:
@ValentinaHutter I opened a PR in the openeo-python-client docs to mention this!

PR #244, opened 2024-04-05
Add basic handling for X,Y,band,time dimensions for transpose and rep…

jzvolensky:
Hi all,

Corresponding issue which this will cover as well: https://github.com/Open-EO/openeo-processes-dask/issues/243.

I am currently working on adding the `align` parameter to the resample_spatial function as per the OpenEO processes [documentation](https://processes.openeo.org/#resample_spatial).

Part of this also involves handling cubes with different dimensions setups. For now, I edited the original fixed transpose to check the shape in the 4 basic dimensions.

Currently I added the `align` as a dummy parameter which does not do anything but allows us to use the function.

Couple of questions and ideas:
1.  Regarding the transpose/handling of the cube dimensions, how could we go about handling other dimensions since there are a lot of possibilities? 
2. How should the align parameter be implemented? what should the result be by using different align options?
3. Since we already use ODC reprojection there is parameter `anchor` available which appears to do a similar operation to align [link to the docs](https://odc-geo.readthedocs.io/en/latest/_api/odc.geo.xr.ODCExtensionDa.reproject.html)

Any feedback, ideas and suggestions are welcome! 

Thanks!

### DIFF
--- a/openeo_processes_dask/process_implementations/cubes/resample.py
+++ b/openeo_processes_dask/process_implementations/cubes/resample.py
@@ -1,7 +1,7 @@
 import logging
 from typing import Optional, Union
 
-import odc.geo.xr 
+import odc.geo.xr
 import rioxarray  # needs to be imported to set .rio accessor on xarray objects.
 from odc.geo.geobox import resolution_from_affine
 from pyproj.crs import CRS, CRSError
@@ -50,7 +50,7 @@ def resample_spatial(
             f'Selected resampling method "{method}" is not available! Please select one of '
             f"[{', '.join(resample_methods_list)}]"
         )
-        
+
     dims = list(data.dims)
 
     known_dims = [
--- a/openeo_processes_dask/process_implementations/cubes/resample.py
+++ b/openeo_processes_dask/process_implementations/cubes/resample.py
@@ -1,9 +1,9 @@
 import logging
-from typing import Optional, Union, Literal
+from typing import Optional, Union
 
-import odc.geo.xr
+import odc.geo.xr 
 import rioxarray  # needs to be imported to set .rio accessor on xarray objects.
-from odc.geo.geobox import resolution_from_affine, AnchorEnum, XY
+from odc.geo.geobox import resolution_from_affine
 from pyproj.crs import CRS, CRSError
 
 from openeo_processes_dask.process_implementations.data_model import RasterCube
@@ -41,8 +41,6 @@ def resample_spatial(
 ):
     """Resamples the spatial dimensions (x,y) of the data cube to a specified resolution and/or warps the data cube to the target projection. At least resolution or projection must be specified."""
 
-    #TODO: Actually implement align parameter in some way
-
     # Assert resampling method is correct.
     if method == "near":
         method = "nearest"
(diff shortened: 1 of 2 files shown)


clausmichele:
It would be good to have feedbacks from the VITO team (@jdries @soxofaan) about how they implemented `resample_spatial` on their side. Do you handle N-dimensional datacubes? Like x,y,bands,t + other dimensions that can be added using `add_dimension`?

jdries:
we don't do adding other dimensions, resample spatial just acts upon the x and y dimensions. The t and bands dimension are of course supported and basically left untouched.

clausmichele:
I didn't mean that you add dimensions within this process, but I was wondering if you support as input datacubes with arbitrary number of dimensions (additional to x and y), which seems so from your reply! Thanks

jdries:
to be clear: our backend does not at all support adding arbitrary number of dimensions. 
We mainly support these combinations for raster cubes:
x,y
x,y,bands
x,y,bands,t
x,y,t

openeo_processes_dask/process_implementations/cubes/resample.py
```diff
@@ -49,14 +52,40 @@ def resample_spatial(
...
+            data_cp = data.transpose('time', 'y', 'x')
+        elif 'band' in dims and 'y' in dims and 'x' in dims:
+            # no time dimension
+            data_cp = data.transpose('band', 'y', 'x')
+        else:
+            # no band and time dimension
+            data_cp = data.transpose('y', 'x')
+    except Exception: 
+        raise OpenEOException(
+        f"Data cube has unexpected dimensions: {dims}"
```
ValentinaHutter, line 77: Thanks for your inputs on the process and adding the align parameter!
I think, we can keep using the `data.openeo` here, as this checks the dimensions in the dataset - and for example uses `t` instead of `time` if the temporal dimension is called like that.

jzvolensky: Hi, sure but using data.openeo limits you to using `band` , `time`, `x` and `y` only now no? Because the moment you add a dimension and run resample it fails. We are looking into ways to extend the support of other dimensions. We can certainly reuse data.openeo for something but ultimately extend the `resample_spatial` dim support.

jzvolensky: we could do something like this maybe:

```Python
    dims = list(data.dims)

    known_dims = [
        data.openeo.band_dims[0],
        data.openeo.temporal_dims[0],
        data.openeo.y_dim,
        data.openeo.x_dim,
    ]

    other_dims = [dim for dim in dims if dim not in known_dims]

    data_cp = data.transpose(*other_dims, *known_dims)
```

This way we can keep the `data.openeo` features plus add other dimensions. Of course, some limits and checks would need to be performed on the `other` dimensions as we cannot realistically allow and support everything.

jzvolensky: Hi Valentina!

I updated the latest commit based on your suggestion to keep the `data.openeo` functionality. The dummy align is set as well to prevent errors from occurring. Let me know what you think!

ValentinaHutter: LGTM! 

The tests are currently failing because you need to `pre-commit run --all-files` on the branch.

ValentinaHutter:
Thanks for the contribution! Will add this to the latest release!

PR #246, opened 2024-04-08
add load_url draft

ValentinaHutter:


ValentinaHutter:
Note: still need to add tests, still need to update output vector cube - prefer xvec, instead of geopandas geodataframe.

PR #249, opened 2024-05-09
resample_spatial missing dimensions causing a crash

jzvolensky:
If dimensions are missing the resample_spatial crashes as it is out of range.

Fix + updated test is on the way


jzvolensky:
Latest commit should fix this issue, @clausmichele @ValentinaHutter can you please review the test as I am not 100% sure it works the way I want it to. It should basically test the different combinations `(x,y,bands)` , `(x,y,t)`, `(x,y)` with the parametrization. If that is not the case I can review it again. Thank you!

clausmichele (approved):

clausmichele:
Hi @jzvolensky, I've reviewed the code and it is fine for me. Please change the status of the PR from draft to ready.

clausmichele:
@jzvolensky tests are not passing anymore. From a quick check it seems for that for Python 3.9 the issue seems related to the odc package. For python 3.10 and 3.11 the error seems to be in how you are calling the `reduce_dimension` (check the reduce_dimension tests to see how to call it).

ValentinaHutter:
We have been looking into the `python3.9` issue this week and decided to remove `python3.9` from the supported versions, Please pull from the latest version. :)

Issue #235, opened 2024-03-04
`load_stac`: possible improvements


clausmichele:
Currently `load_stac` relies on the [stackstac](https://github.com/gjoseph92/stackstac) library, which works fine until we try to load more complex types of STAC objects.

As far as I am aware, current limitations of stackstac are:
- Can load only single band COGs: multi-band geoTIFFs are not supported
- Can't load STAC based on netCDF data

This limits the functionality of the implemented process. I will investigate the possible switch to either:
- [odc-stac](https://github.com/opendatacube/odc-stac): regularly maintained and updated
- [xpystac](https://github.com/stac-utils/xpystac): based on both `stackstac` and `odc-stac`

clausmichele:
@jonas-eberle we created a new updated PR here: https://github.com/Open-EO/openeo-processes-dask/pull/328

PR #284, opened 2024-10-02
`load_stac` enhancements

clausmichele:
This PR will include many enhancements for the `load_stac` process:

- [x] Switch from `stackstac` to `odc.stac`, see the reasons here: https://github.com/Open-EO/openeo-processes-dask/issues/235
- [x] Increase support for other data types. With this PR, it's possible to load Zarr data, see this example: https://gist.github.com/clausmichele/ba9fdbf5add71504b7553f7f7b6d1be0
- [x] Maintain the source data type instead of converting everything to float64, reducing the memory and computational resources required.
- [ ] Get the dimension names from the STAC datacube extension, if available (still not sure about this, need to discuss)

If you have any other suggestion of possible improvements, please comment


clausmichele:
Closing in favor of github.com/Open-EO/openeo-processes-dask/pull/328

Issue #257, opened 2024-06-05
`resample_cube_spatial` should not crop the data spatially


clausmichele:
See here: https://github.com/Open-EO/openeo-processes/issues/509

ValentinaHutter:
https://github.com/Open-EO/openeo-processes-dask/pull/300

clausmichele:
@ValentinaHutter the latest implementation of resample_cube_spatial breaks our notebook, could you please have a look? The returned error is: `Exception: Coordinates could not be aligned! `


but the process is designed exactly to align different resolutions/projections.

https://github.com/EO-College/cubes-and-clouds/blob/main/lectures/2.3_data_access/exercises/23_data_access_resample.ipynb

ValentinaHutter:
I now created a new PR to fix this: https://github.com/Open-EO/openeo-processes-dask/pull/314

I used the following code to test the process directly in openeo-processes-dask, but will not merge it in, because for these tests, we do not want to depend on an external stac collection: 

```
import pytest
from openeo_pg_parser_networkx.pg_schema import TemporalInterval, BoundingBox

from openeo_processes_dask.process_implementations.cubes.resample import (
    resample_cube_spatial,
)
from openeo_processes_dask.process_implementations.cubes.load import load_stac


def test_resample_cube_spatial(
):
    url = "https://earth-search.aws.element84.com/v1/collections/sentinel-2-l2a"

    spatial_extent = BoundingBox.model_validate({"west": 11.4, "east": 11.42, "south": 45.5, "north": 45.52})
    temporal_extent = TemporalInterval.model_validate(["2023-06-01", "2023-06-30"])
    bands = ["red","nir"]

    s2_cube = load_stac(url=url,
        spatial_extent=spatial_extent,
        temporal_extent=temporal_extent,
        bands=bands
        )
    url = "https://planetarycomputer.microsoft.com/api/stac/v1/collections/landsat-c2-l2"
    bands = ["red","nir08"]
    l8_cube = load_stac(url=url,
        spatial_extent=spatial_extent,
        temporal_extent=temporal_extent,
        bands=bands)

    
    output_cube = resample_cube_spatial(
        data=l8_cube, target=s2_cube, method="average"
    )
    print(s2_cube[s2_cube.openeo.x_dim[0]].values == output_cube[output_cube.openeo.x_dim[0]].values)

    output_cube = resample_cube_spatial(
        data=s2_cube, target=l8_cube, method="average"
    )
[...]

PR #300, opened 2024-11-26
Resample cube update

ValentinaHutter:


ValentinaHutter:
Thanks for the feedback! I just reopened the issue! Will check today! https://github.com/Open-EO/openeo-processes-dask/issues/257

PR #308, opened 2025-01-07
first implementation of local udf

VincentVerelst:


clausmichele:
@VincentVerelst there seem to be conflicts. Could you fix them before proceeding?

VincentVerelst:
@clausmichele , merge conflicts have been solved

openeo_processes_dask/process_implementations/udf/udf.py
```diff
@@ -0,0 +1,26 @@
...
+from openeo_processes_dask.process_implementations.data_model import RasterCube
+
+__all__ = ["run_udf"]
+
+
+def run_udf(
+    data: da.Array, udf: str, runtime: str, context: Optional[dict] = None
+) -> RasterCube:
+    data = XarrayDataCube(xr.DataArray(data))
+    data = UdfData(proj={"EPSG": 900913}, datacube_list=[data], user_context=context)
```
clausmichele, line 18: Why this default EPSG code?

clausmichele (changes requested):
Some things to fix:

1. There is a missing `__init__.py` file. It has to be inside the `openeo-processes-dask/process_implementations/udf` folder, with the following content:
```python
from .udf import *
```
2. The attributes are not maintained and with them the projection/CRS is also gone. It's necessary to keep it. You can test it with this MWE:

```python
import openeo

temporal_extent = ["2022-06-01", "2022-06-10"]
bands = ["red", "nir"]
    url=url,
    bands=bands,
    properties=properties,
)

b04 = s2_datacube.band("red")
b08 = s2_datacube.band("nir")
ndvi = (b08 - b04) / (b08 + b04)
ndvi_median = ndvi.reduce_dimension(dimension="time", reducer="median")

result_ndvi_xr = ndvi_median.execute()

# Build a UDF object from an inline string with Python source code.
udf = openeo.UDF("""
from openeo.udf import XarrayDataCube

def apply_datacube(cube: XarrayDataCube, context: dict) -> XarrayDataCube:
    array = cube.get_array()
    print(array.shape)
    array.values = 0.0001 * array.values
    return cube
""")

# Apply the UDF to a cube.
rescaled_cube = ndvi_median.apply(process=udf)

rescaled_cube_xr = rescaled_cube.execute()

print(result_ndvi_xr.rio.crs)
print(rescaled_cube_xr.rio.crs)
```

openeo_processes_dask/process_implementations/udf/udf.py
jdries: legacy stuff, better to remove it here

openeo_processes_dask/process_implementations/udf/udf.py
```diff
@@ -0,0 +1,26 @@
...
+
+from openeo_processes_dask.process_implementations.data_model import RasterCube
+
+__all__ = ["run_udf"]
+
+
+def run_udf(
+    data: da.Array, udf: str, runtime: str, context: Optional[dict] = None
+) -> RasterCube:
+    data = XarrayDataCube(xr.DataArray(data))
```
soxofaan, line 17: Shouldn't there also be some kind of chunking here (or at least some kind of option to enable it), to make sure it can be aligned a bit with backends that automatically apply chunking (like geopyspark driver).

(I don't know the dask driver details very well, so chunking might already be in play at an other level)

tests/test_udf.py
```diff
@@ -0,0 +1,38 @@
...
+        spatial_extent=bounding_box,
+        temporal_extent=temporal_interval,
+        bands=["B02", "B03", "B04", "B08"],
+        backend="dask",
+    )
+
+    udf = """
+from openeo.udf import XarrayDataCube
+def apply_datacube(cube: XarrayDataCube, context: dict) -> XarrayDataCube:
+    return cube
```
soxofaan, line 25: Isn't this UDF too simple to actually be able verify that it did something?
E.g. I would at least the values with 2 or 10 or -1 so you can actually verify the UDF ran.

openeo_processes_dask/process_implementations/udf/udf.py
clausmichele: Indeed, but as far as I understood @jdries and @VincentVerelst wanted to have a first implementation ready soon and later on focus on the chunking part.

tests/test_udf.py
```diff
@@ -0,0 +1,38 @@
...
+        data=random_raster_data,
+        spatial_extent=bounding_box,
+        temporal_extent=temporal_interval,
+        bands=["B02", "B03", "B04", "B08"],
+        backend="dask",
+    )
+
+    udf = """
+from openeo.udf import XarrayDataCube
+def apply_datacube(cube: XarrayDataCube, context: dict) -> XarrayDataCube:
```
jdries, line 24: shall we use the signature that immediately takes an xarray.DataArray? Would only be to make the test use things that we would recommend in practice.

openeo_processes_dask/process_implementations/__init__.py
```diff
@@ -9,6 +9,7 @@
 from .logic import *
 from .math import *
 from .text import *
+from .udf import *
```
jdries, line 12: It seems this works better if we want to get run_udf in the ProcessRegistry
```suggestion
from .udf.udf import *
```

openeo_processes_dask/process_implementations/udf/udf.py
jdries: Correct, I did now try it locally, and it seems that it also works as a callback to an apply. 
I do see that chunking is not necessarily present, but think that is caused by having a very small input array to begin with. I would propose to continue with wrapping up this one and then indeed add the more advanced tests.

openeo_processes_dask/process_implementations/__init__.py
soxofaan: I guess this is related to the remark of @clausmichele  that there is a `__init__.py` missing in the `udf` folder

VincentVerelst:
> The attributes are not maintained and with them the projection/CRS is also gone. It's necessary to keep it. You can test it with this MWE:

This seems to be a broader problem in the `apply` process. Seems to be solved by adding `keep_attrs=True` in `apply_ufunc`.

clausmichele:
> > The attributes are not maintained and with them the projection/CRS is also gone. It's necessary to keep it. You can test it with this MWE:
> 
> This seems to be a broader problem in the `apply` process. Seems to be solved by adding `keep_attrs=True` in `apply_ufunc`.

@ValentinaHutter do you know if the attributes were not passed due to some technical reason? Maybe some edge case that would not work if we keep the attributes?

ValentinaHutter:
> > > The attributes are not maintained and with them the projection/CRS is also gone. It's necessary to keep it. You can test it with this MWE:
> > 
> > 
> > This seems to be a broader problem in the `apply` process. Seems to be solved by adding `keep_attrs=True` in `apply_ufunc`.
> 
> @ValentinaHutter do you know if the attributes were not passed due to some technical reason? Maybe some edge case that would not work if we keep the attributes?

It should be fine to set `keep_attrs=True`, there is no edge case, I know of... I think it was just not added before because we did not depend on the attributes and did not experience related issues.

VincentVerelst:
@clausmichele , two unit tests are failing because `dask_geopandas` 0.4.3 was released. `GeoDataFrame` is now part of `dask_geopandas.expr` instead of `dask_geopandas.core`. So we either change the dask_geopadas dependency to <0.4.3, or we change this line: https://github.com/Open-EO/openeo-processes-dask/blob/main/openeo_processes_dask/process_implementations/ml/random_forest.py#L98

Let me know what you prefer.

clausmichele:
> @clausmichele , two unit tests are failing because `dask_geopandas` 0.4.3 was released. `GeoDataFrame` is now part of `dask_geopandas.expr` instead of `dask_geopandas.core`. So we either change the dask_geopadas dependency to <0.4.3, or we change this line: https://github.com/Open-EO/openeo-processes-dask/blob/main/openeo_processes_dask/process_implementations/ml/random_forest.py#L98
> 
> Let me know what you prefer.

I'm not the author of that function and we are not using it at the moment. EODC developed it and probably @ValentinaHutter knows if the change could impact something on their side.

ValentinaHutter:
thanks for adding the change to 0.4.3! from EODC's side it's fine to switch to this version.

VincentVerelst:
@ValentinaHutter , @clausmichele , @jdries , how do we proceed further?

clausmichele (approved):

ValentinaHutter:
Looks like tests are failing because of a change in either the test data or a dependency in load_stac. I now created a PR to independently test the ndvi process https://github.com/Open-EO/openeo-processes-dask/pull/316 but for `load_stac`, the issue still needs to be resolved.

clausmichele:
I will investigate! Something changed but I don't understand where. In fact, even the example in the docs now is failing: https://open-eo.github.io/openeo-python-client/cookbook/localprocessing.html

clausmichele:
So, the issue occurs with the latest version of pystac. With `pystac==1.9.0` it doesn't occur. @ValentinaHutter could you fix `pystac==1.9.0` in the project?

ValentinaHutter:
Did you mean pystac or pystac-client?

clausmichele:
`pystac`, currently it is a dependency of stackstac (and also odc-stac if we will use it in the near future) and we don't have it in the requirements.

ValentinaHutter:
Thanks for the clarification, I now merged the dependency into main, so you can pull from there :+1:

clausmichele:
@VincentVerelst if you merge the latest `pystac` addition into `pyproject.toml` it will fix the tests and we can proceed with the merge.

VincentVerelst:
@clausmichele , done!

HansVRP:
@clausmichele can we merge this one, and are there any follow-up tickets we could contribute to?

clausmichele:
> @clausmichele can we merge this one, and are there any follow-up tickets we could contribute to?

For me yes, but I don't have the rights to merge PRs. Only EODC has admin rights here... @ValentinaHutter @GeraldIr @SerRichard @christophreimer

PR #314, opened 2025-01-17
roll back resample cube spatial

ValentinaHutter:


clausmichele:
Hi @ValentinaHutter, the current resample_cube_spatial implementation is still broken. I wonder if it wouldn't be more efficient to roll back to the previous implementation in the meantime.

PR #322, opened 2025-02-19
run_udf: add tests for case inside callback

jdries:
These tests show how a user would typically use run_udf.

They also seem to expose a remaining issue in current implementation, in the sense that the UDF continues to receive the full input datacube as opposed to only receiving a chunk.

### DIFF
--- a/tests/test_udf.py
+++ b/tests/test_udf.py
@@ -1,9 +1,18 @@
+from functools import partial
+
 import numpy as np
 import openeo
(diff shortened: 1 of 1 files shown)


ValentinaHutter:
Pre-commit should be included in your poetry setup, but you can also resolve the pre-commit hook issue manually by running: `pre-commit run --all-files` 

https://pre-commit.com/

PR #326, opened 2025-05-12
First implementation of the oscar_python client

jzvolensky:
process json PR: https://github.com/eodcgmbh/openeo-processes/pull/46

adds oscar_client and minio as dependencies

implements the oscar_python client library to interact with an OSCAR service

There will probably be a bunch of small changes and fixes here in there as it is quite extensive

CC @clausmichele 

@SerRichard is the current `_get_refresh_token` enough for what we discussed about reading the token from an env variable


clausmichele:
From what I remember of our discussion, the token should be handled by the openEO back-end and we should "just" read it from an environment variable. Please check my review of the proposed process definition.

PR #329, opened 2025-07-04
Add enhanced filter_bbox with VectorCube support

Yuvraj198920:
- Added VectorCube support to filter_bbox function
- Fixed critical coordinate transformation bug in _reproject_bbox
- Improved geometric filtering using intersects() instead of within()
- Enhanced spatial filtering for both RasterCube and VectorCube types
- Maintains backward compatibility with existing functionality

Fixes:
- Lines 280-310: Corrected bbox coordinate order and transformer parameters
- Lines 267, 273: Changed geometric filtering to include boundary cases
- Line 173: Updated return type annotation for Union support


jzvolensky:
Note: the included examples are temporary and will be removed from the PR

clausmichele:
@Yuvraj198920 you should remove all the files added in the root folder and add a test under `./tests/` compliant with pytest.

jzvolensky:
@ValentinaHutter Hello may I ask you to please review this and see if it can be merged? Thank you very much!

clausmichele:
@Yuvraj198920 I see that you added a `filter_bbox.py`, but the changes should go in the `_filter.py` https://github.com/Open-EO/openeo-processes-dask/blob/main/openeo_processes_dask/process_implementations/cubes/_filter.py

Yuvraj198920:
@clausmichele hi, added test for filter_bbox_vectorcube.

ValentinaHutter (commented):
Is the `tests/data/Netherlands_polygon.geojson` required? I do not see it being used in the test? 

Also, it would be good to include a test with a different crs being used to see if the `_reproject_bbox` works as expected. 

Thanks a lot for these updates!

Yuvraj198920:
Hi @ValentinaHutter
- Removed tests/data/Netherlands_polygon.geojson (unused file)
- Added test_filter_bbox_vectorcube_crs_reprojection() to validate _reproject_bbox function
- Test uses VectorCube in UTM Zone 32N with WGS84 bbox to ensure CRS transformation works correctly

PR #332, opened 2025-07-18
Fix spatial dimension alignment for float coordinates in merge_cubes

suriyahgit:
### Description of Changes:

**merge.py changes:**
1. Added a new constant `FLOAT_TOLERANCE = 1e-6` to define the acceptable difference between float coordinates
2. Implemented a new helper function `_align_coordinates()` that:
   - Identifies shared dimensions between cubes
   - Checks if coordinates are floating-point numbers
   - Verifies coordinate shapes match
   - Aligns coordinates if they differ by less than `FLOAT_TOLERANCE`
3. Modified `merge_cubes()` to call `_align_coordinates()` before processing
4. Added CRS preservation when reducing with overlap_resolver

**Test changes:**
1. Added new test case `test_merge_float_coord_alignment` in the `tests/test_merge.py` that:
   - Creates two identical cubes with slightly different float coordinates (differing by 1e-6)
   - Verifies they can be successfully merged
   - Checks the output is a valid xarray DataArray

**Purpose of changes:**
The changes address an edge case where cubes with nearly identical floating-point coordinates (due to minor numerical differences) would fail to merge. This commonly occurs with spatial coordinates that might have tiny differences due to different processing pipelines or floating-point arithmetic.

The solution:
1. Automatically aligns coordinates that are very close numerically
2. Preserves the original cube's coordinates when they differ within tolerance
3. Maintains all existing functionality for non-float coordinates and larger differences
4. Ensures CRS information is preserved during merge operations

The test verifies this behavior by creating cubes with intentionally misaligned coordinates (within tolerance) and confirming they can be merged successfully.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,6 +16,7 @@ on:
 env:
   POETRY_VERSION: 1.5.1
 
+
 jobs:
   tests:
     name: ${{ matrix.os }} / ${{ matrix.python-version }}
@@ -23,10 +24,10 @@ jobs:
     strategy:
       matrix:
         os: [Ubuntu]
-        python-version: ["3.10", "3.11"]
+        python-version: ["3.10", "3.11", "3.12"]
         include:
           - os: Ubuntu
-            image: ubuntu-22.04
+            image: ubuntu-24.04   # switch to Noble (24.04)
       fail-fast: false
     defaults:
       run:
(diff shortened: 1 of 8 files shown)


clausmichele:
Hi @ValentinaHutter, this PR is ready for review. The tests are failing on `test_ddmc_target_band`, which is not related with this PR.

ValentinaHutter:
Thanks for the great work on that! I just had a look into what is causing the issue, and it seems that `.assign_coords` removes the `openeo.band_dims` information. So, this could also cause trouble when using `.add_dimension` or `.rename_dimension` which would allow for `openeo.band_dims` other than the expected ones. It seems like using `cube2[dim] = ` passes the `openeo.band_dims` through, so I added the suggestion to also fix the test.

suriyahgit:
Hi @ValentinaHutter, this PR now passes the gitCI, kindly review it, thanks!

PR #335, opened 2025-09-04
Draft: add rqatrend

ValentinaHutter:


openeo_processes_dask/process_implementations/experimental/rqadeforestation.py
```diff
@@ -0,0 +1,40 @@
...
+                ("s1", ct.c_int64)]
+
+def mvptr(A):
+    ptr = A.ctypes.data_as(ct.c_void_p)
+    a = MallocVector(ptr, ct.c_int64(A.size), ct.c_int64(A.shape[0]))
+    return ct.byref(a)
+
+
+# download so file at https://github.com/EarthyScience/RQADeforestation.py/archive/refs/heads/main.zip
+lib = ct.CDLL("./.venv/lib/python3.11/site-packages/rqadeforestation/lib/rqatrend.so")
```
danlooo: This path probably depends on the environment. The so file will be downloaded during install, see [pyproject.toml](https://github.com/EarthyScience/RQADeforestation.py/blob/4c0af502e59e999e2a7583f425b5fcce0b3534b2/pyproject.toml#L25). Shall I move the lib object and the malloc/pointer functions to rqadeforestation.py?

pyproject.toml
```diff
@@ -47,6 +47,7 @@ openeo = ">=0.36.0"
 numpy = { version = "<2.0.0", optional = false }
 pystac = { version = "<1.12.0", optional = false }
 zarr = "<=2.18.7"
+rqadeforestation = {git = "https://github.com/EarthyScience/RQADeforestation.py"}
```
danlooo, line 50: Is there a way to use a specific git ref to ensure reproducibility? We now have tag v0.1.

PR #336, opened 2025-09-10
Add rqatrend 

danlooo:
This PR aims to add the deforestation detection workflow of the [FAIRSenDD](https://github.com/EarthyScience/FAIRSenDD) project by using its python bindings [rqadeforestation.py](https://github.com/EarthyScience/rqadeforestation.py). 

- continues work of #335 
- closes #335
- see also https://github.com/eodcgmbh/openeo-processes/pull/49

### DIFF
--- a/openeo_processes_dask/process_implementations/experimental/__init__.py
+++ b/openeo_processes_dask/process_implementations/experimental/__init__.py
@@ -1,1 +1,2 @@
 from .ddmc import *
+from .rqadeforestation import *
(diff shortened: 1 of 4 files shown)


danlooo:
@ValentinaHutter Do we really need the mechanism of  `ct.CDLL( f"./.venv/lib/python{version}"` ?  The so files and lib object are already part of the python binding package, see https://github.com/EarthyScience/RQADeforestation.py/blob/4c0af502e59e999e2a7583f425b5fcce0b3534b2/rqadeforestation/__init__.py#L28
Is there an environment so we can test this?

PR #354, opened 2025-11-21
Refactor GDAL installation to micromamba (conda-forge) and add xcube tests

suriyahgit:
### **PR Description**

This PR improves the stability and reproducibility of the environment used in CI by moving GDAL and related scientific dependencies away from `apt`-based installation and into a controlled micromamba environment sourced from `conda-forge`.

### **Key changes**

* Removed GDAL and GIS dependency installation via `apt`
* Added GDAL installation within micromamba (`gdal=3.11.*`, strict channel priority)
* Updated CI workflow to mimic the dependency handling approach used in the openEO backend
* Added xcube-related tests under the `tests` folder
* Ensures compatibility with Python 3.10, 3.11 and 3.12

### **Motivation**

`apt` GDAL packages are tied to OS release cycles and caused breakages when testing with newer Python versions (e.g., Python 3.12 + GDAL 3.6.2). Installing GDAL from `conda-forge` provides stable builds and avoids needing to update Docker/CI configurations frequently.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -72,15 +72,6 @@ jobs:
           poetry config virtualenvs.create false
           poetry config virtualenvs.in-project false
 
-      # (Optional but often handy) cache Poetry's virtualenv-less install
-      - name: Cache Poetry dependencies
-        uses: actions/cache@v3
-        with:
-          path: ~/.cache/pypoetry
-          key: poetry-${{ runner.os }}-${{ matrix.python-version }}-${{ hashFiles('**/poetry.lock') }}
-          restore-keys: |
-            poetry-${{ runner.os }}-${{ matrix.python-version }}-
-
       - name: Install project dependencies with Poetry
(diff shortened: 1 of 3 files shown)


suriyahgit:
Poetry was implicitly resolving `gdal` from PyPI and upgrading it to `3.12.0.post1`, which conflicted with `libgdal 3.11.*` (`gdal=3.11.*`) installed via micromamba in CI, causing build failures. Explicitly pinning:

```toml
gdal = ">=3.11,<3.12"
```
ensures compatibility with the conda-forge GDAL version and prevents accidental resolver upgrades.

ValentinaHutter:
Would it be possible for you to re-trigger the tests?
In my setup, I cannot install gdal 3.11 anymore, since 3.12 is already released and 3.11 is no longer available. This will happen with 3.12 as soon as 3.13 is released as well, so we cannot really pin gdal to a specific version.

PR #372, opened 2026-04-10
Modernize dependencies and CI, add Python 3.13 support, and fix NumPy 2 compatibility issues

suriyahgit:
This PR modernises `openeo-processes-dask` dependencies and CI, extends Python support to 3.13, restores compatibility with current versions of NumPy, XGBoost, and pystac, and fixes a set of runtime and test issues uncovered during the upgrade work. Removes the xcube-eopf dependency and EOPF data support in `load_stac`.

The main goals are:

* allow newer dependency versions without dropping existing supported lower bounds where possible
* remove deprecated NumPy and Pydantic usage
* make CI and release workflows more robust by replacing the UbuntuGIS PPA setup with conda-forge
* fix correctness and compatibility issues in `array_find`, `array_interpolate_linear`, `eq`, `resample_cube_temporal`, and `random_forest`

---

### Python support

* Supported Python range updated from `>=3.10,<3.13` to `>=3.10,<3.14`
* Added Python 3.13 and 3.14 classifiers

### Package version

* `openeo-processes-dask`: `2025.10.1` → `2026.06.2`

### Dependency policy

This PR follows a “widen, don’t shift” approach: upper bounds are removed where possible to allow newer versions, while existing lower bounds are largely preserved to avoid unnecessary disruption for older environments.

### Updated constraints

| Package                             | Old                       | New           |
| ----------------------------------- | ------------------------- | ------------- |
| `numpy`                             | `<2.0.0`                  | `>=1.26.3,<3` |
| `xarray`                            | `>=2022.11.0,<2025.08.01` | `>=2022.11.0` |
| `dask[array,dataframe,distributed]` | `>=2023.4.0,<2025.2.0`    | `>=2023.4.0`  |
| `xgboost`                           | `>=1.5.1,<2.1.4`          | `>=1.5.1`     |
| `pystac`                            | `<1.12.0`                 | `>=1.8.0`     |
| `geopandas`                         | `>=0.11.1,<1`             | `>=0.11.1`    |
| `dask-geopandas`                    | `0.4.3`                   | `>=0.4.3`     |
| `pyarrow`                           | `^15.0.2`                 | `>=15.0.2`    |
| `openeo-pg-parser-networkx`         | `>=2025.10`               | `>=2025.10.0` |

### New optional dependency

* `gdal >=3.9` added as an explicit optional dependency, installed from conda-forge in CI

---

## CI and release workflow changes

The CI and release workflows were reworked to install GDAL via conda-forge instead of the UbuntuGIS PPA.

### Main changes

* switched GDAL installation from UbuntuGIS PPA to Miniforge/mamba + conda-forge
* added `.github/ci-environment.yml` to define the CI conda environment
* configured Poetry to install into the conda environment with `virtualenvs.create false`
* expanded the test matrix from Python 3.10–3.12 to 3.10–3.13
* upgraded Poetry from `1.5.1` to `2.3.3`
* updated GitHub Actions versions
* changed shell from `bash` to `bash -l {0}` for conda activation
* replaced the old `.venv` cache strategy with separate caches for conda packages and pip wheels
* added the missing `poetry install` step in the release workflow
* moved workflow commands to run directly inside the conda environment instead of through `poetry run`

This removes a fragile CI path that depended on matching the Python GDAL binding to the system GDAL version at runtime.

### New file

`.github/ci-environment.yml`

```yaml
name: ci
channels:
  - conda-forge
dependencies:
  - gdal
```

---

#### `utils.py` — `get_scalar_type()`

Replaced `np.obj2sctype()` with logic based on `np.asarray(obj).dtype.type` and `np.dtype(obj.dtype).type`, which works on both NumPy 1.x and 2.x.

Also added a `str` fast path so string inputs are handled consistently.

#### `comparison.py` — `np.issubsctype()` removal

Replaced all uses of `np.issubsctype()` with `np.issubdtype()`.

In `eq()`, the incompatible-type branch now returns a boolean array matching input shape instead of a scalar `False`, avoiding shape-related issues for array inputs.

---

#### `array_find()`

Reworked `array_find()` to fix three issues:

* `np.isnan(value)` could fail on non-numeric values
* masked-array behaviour was unreliable for Dask-backed results under NumPy 2.x because masks were lost when coerced
* the function emitted a warning on every call about masked arrays

The new behaviour is:

* handle non-numeric search values safely
* return filled arrays instead of masked arrays
* keep the Dask path lazy
* remove the per-call warning

For not-found positions, the function now returns filled values (`999999`) instead of masked entries.

#### `array_interpolate_linear()`

Fixed the Dask 1D path so it stays lazy by using `da.map_blocks(...)`.

Also:

* converts block inputs to NumPy arrays inside the interpolation helper
* avoids mutating the caller input in place by copying before assignment
* simplifies `valid.all()` logic

---

### `random_forest.py`

Modernised XGBoost Dask usage by switching from `xgb.dask.*` access through `import xgboost as xgb` to:

```python
from xgboost import dask as dxgb
```

Also added `ImportError` guards with clearer error messages.

---

#### `resample_cube_temporal()`

Simplified tie handling when multiple timestamps are equally close to the target.

The old implementation only handled specific `argwhere()` output shapes. The new implementation flattens the index result and always takes the first match, which consistently selects the earliest timestamp and avoids crashes on other tie configurations.

---

### `cubes/apply_neighborhood_intertwin.py`

Added an explicit zero-size dimension check and now raises a clear `ValueError` when an input dimension has size 0.

---

### Pydantic v2

Replaced all `parse_obj()` uses with `model_validate()` across fixtures and tests.

### Test fixes

* fixed a Dask assertion that previously passed without actually evaluating values by adding `.compute()`
* corrected the `array_find()` string not-found expectation from `99999` to `999999`

The previous string sentinel test was not validating the actual returned value because masked positions were effectively being ignored.

---

## Behavioural changes to note

These are the main observable behaviour changes introduced by the fixes:

1. `numpy >=1.26.3` is now required, and NumPy 2.x is supported.

2. `pystac >=1.8.0` is now required.

3. `array_find()` now returns filled `np.ndarray` / `da.Array` results instead of masked arrays. Not-found positions are represented by `999999`.

4. `array_interpolate_linear()` now preserves laziness for 1D Dask inputs instead of forcing eager behaviour.

5. `eq()` now returns a shape-preserving boolean array for incompatible array/type comparisons instead of a scalar `False`.

6. CI now assumes Poetry 2.x compatibility.

---

## Testing notes

* CI now runs on Python 3.10, 3.11, 3.12 and 3.13
* GDAL is installed from conda-forge
* corrected tests now validate previously untested behaviour in `eq()` and `array_find()`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -48,6 +48,7 @@ jobs:
           channel-priority: strict
           environment-file: .github/ci-environment.yml
           use-mamba: true
+          conda-remove-defaults: true
 
       # Cache downloaded conda tarballs (NOT the resolved environment)
       - name: Cache conda package downloads
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,12 +33,12 @@ jobs:
       run:
         shell: bash
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           submodules: recursive
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v6
         with:
           python-version: ${{ matrix.python-version }}
 
@@ -66,7 +66,7 @@ jobs:
         run: poetry config virtualenvs.in-project true
 
       - name: Set up cache
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         id: cache
         with:
           path: .venv
@@ -91,6 +91,6 @@ jobs:
         run: poetry run pytest --cov=openeo_processes_dask --cov-report=xml
 
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v4
+        uses: codecov/codecov-action@v6
         with:
           files: ./coverage.xml
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -45,7 +45,7 @@ jobs:
           activate-environment: ci
           channels: conda-forge
           channel-priority: strict
-          environment-file: ci-environment.yml
+          environment-file: .github/ci-environment.yml
           use-mamba: true
 
       # Cache the conda env + package cache
@@ -56,7 +56,7 @@ jobs:
           path: |
             ~/conda_pkgs_dir
             ${{ env.CONDA }}/envs/ci
-          key: conda-${{ runner.os }}-py${{ matrix.python-version }}-${{ hashFiles('ci-environment.yml') }}
+          key: conda-${{ runner.os }}-py${{ matrix.python-version }}-${{ hashFiles('.github/ci-environment.yml') }}
           restore-keys: |
             conda-${{ runner.os }}-py${{ matrix.python-version }}-
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,8 +14,8 @@ on:
   workflow_dispatch:
 
 env:
-  POETRY_VERSION: 1.5.1
-
+  POETRY_VERSION: 2.3.3
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
 
 jobs:
   tests:
@@ -27,18 +27,18 @@ jobs:
         python-version: ["3.10", "3.11", "3.12"]
         include:
           - os: Ubuntu
-            image: ubuntu-24.04   # switch to Noble (24.04)
+            image: ubuntu-24.04
       fail-fast: false
     defaults:
       run:
         shell: bash
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           submodules: recursive
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
 
@@ -48,10 +48,8 @@ jobs:
           sudo apt-get install -y software-properties-common
           sudo add-apt-repository ppa:ubuntugis/ubuntugis-unstable -y
           sudo apt-get update
-          # Install GDAL from the PPA (brings matching libgdal)
           sudo apt-get install -y libgdal-dev gdal-bin
           gdalinfo --version
-          # Show where libgdal comes from (should be /usr/lib/* from UbuntuGIS)
           apt-cache policy libgdal-dev | sed -n '1,20p'
 
       - name: Get full Python version
@@ -68,7 +66,7 @@ jobs:
         run: poetry config virtualenvs.in-project true
 
       - name: Set up cache
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         id: cache
         with:
           path: .venv
@@ -80,12 +78,10 @@ jobs:
           [ "$(command -v timeout)" ] || function timeout() { perl -e 'alarm shift; exec @ARGV' "$@"; }
           timeout 10s poetry run pip --version || rm -rf .venv
 
-      # Important: make Poetry resolve GDAL against system libgdal=3.11
-      # If you explicitly pin GDAL in pyproject, set it to "==3.11.3".
       - name: Install dependencies
         run: |
           ver="$(gdal-config --version)"
-          poetry add "gdal==${ver}"          # adds to main group
+          poetry add "gdal==${ver}"
           poetry install --with dev --all-extras
 
       - name: Pre-commit hooks
@@ -95,6 +91,6 @@ jobs:
         run: poetry run pytest --cov=openeo_processes_dask --cov-report=xml
 
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v3
+        uses: codecov/codecov-action@v4
         with:
           files: ./coverage.xml
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,64 +30,82 @@ jobs:
       fail-fast: false
     defaults:
       run:
-        shell: bash
+        shell: bash -l {0}
     steps:
       - uses: actions/checkout@v5
         with:
           submodules: recursive
 
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v6
+      # --- Conda env with GDAL from conda-forge (cached) ---
+      - name: Set up Miniforge + Python ${{ matrix.python-version }}
+        uses: conda-incubator/setup-miniconda@v3
         with:
+          miniforge-version: latest
           python-version: ${{ matrix.python-version }}
+          activate-environment: ci
+          channels: conda-forge
+          channel-priority: strict
+          environment-file: ci-environment.yml
+          use-mamba: true
 
-      - name: Install system GDAL 3.11.3 (UbuntuGIS)
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y software-properties-common
-          sudo add-apt-repository ppa:ubuntugis/ubuntugis-unstable -y
-          sudo apt-get update
-          sudo apt-get install -y libgdal-dev gdal-bin
-          gdalinfo --version
-          apt-cache policy libgdal-dev | sed -n '1,20p'
-
-      - name: Get full Python version
-        id: full-python-version
-        run: echo version=$(python -c "import sys; print('-'.join(str(v) for v in sys.version_info))") >> $GITHUB_OUTPUT
-
-      - name: Bootstrap poetry
-        run: curl -sSL https://install.python-poetry.org | python3 - -y --version $POETRY_VERSION
-
-      - name: Update PATH
-        run: echo "$HOME/.local/bin" >> $GITHUB_PATH
-
-      - name: Configure poetry
-        run: poetry config virtualenvs.in-project true
-
-      - name: Set up cache
+      # Cache the conda env + package cache
+      - name: Cache conda environment
         uses: actions/cache@v5
-        id: cache
+        id: conda-cache
         with:
-          path: .venv
-          key: venv-${{ runner.os }}-${{ steps.full-python-version.outputs.version }}-${{ hashFiles('**/poetry.lock') }}
+          path: |
+            ~/conda_pkgs_dir
+            ${{ env.CONDA }}/envs/ci
+          key: conda-${{ runner.os }}-py${{ matrix.python-version }}-${{ hashFiles('ci-environment.yml') }}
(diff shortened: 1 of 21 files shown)


suriyahgit:
also a spiritual continuity from these PRs: https://github.com/Open-EO/openeo-processes-dask/pull/358 (numpy v2 support); https://github.com/Open-EO/openeo-processes-dask/pull/354 and https://github.com/Open-EO/openeo-processes-dask/pull/345 (GDAL related CI stuffs which this PR solves)

suriyahgit:
codecov is comparing against an outdated base commit (`010c214`) that has **no coverage uploads**, so the reported coverage change is not reliable. This likely stems from missing/invalid coverage reports on recent `main` commits, not from the changes in this PR. We should trigger a fresh CI run on main (or merge this PR) to ensure coverage is uploaded with the updated workflow, so Codecov can establish a valid baseline for future comparisons.

openeo_processes_dask/process_implementations/utils.py
```diff
@@ -4,8 +4,10 @@
 
 
 def get_scalar_type(obj):
+    if isinstance(obj, str):
+        return np.str_
     if np.isscalar(obj):
-        return np.obj2sctype(type(obj))
+        return np.asarray(obj).dtype.type
```
pjonsson: https://numpy.org/doc/2.3/numpy_2_0_migration_guide.html says use `np.dtype(obj).type` instead of `obj2sctype`, and NumPy 1.25 seem to contain np.dtype (https://numpy.org/doc/1.25/reference/arrays.dtypes.html). Why is all the extra logic required?

openeo_processes_dask/process_implementations/arrays.py
```diff
@@ -296,28 +296,34 @@ def array_find(
     if reverse:
         data = np.flip(data, axis=axis)
 
-    idxs = (data == value).argmax(axis=axis)
+    eq = data == value
+    idxs = eq.argmax(axis=axis)
+    mask = ~np.array(eq.any(axis=axis))
+
+    # Fix: np.isnan crashes on strings, None, lists, etc.
```
pjonsson: I find it difficult to sort out what fixes that are done because of Python 3.13/3.14 incompatibilities, NumPy 2 incompatibilities, or CI incompatibilities. Is it a NumPy 2 problem that `np.isnan` crashes on anything that isn't `array_like`?

pjonsson:
> This likely stems from missing/invalid coverage reports on recent `main` commits, not from the changes in this PR. We should trigger a fresh CI run on main (or merge this PR) to ensure coverage is uploaded with the updated workflow, so Codecov can establish a valid baseline for future comparisons.

Does code coverage work for main, but CodeCov was down when the last commit was merged to main? If so, merging #358 and rebasing this on top of latest main should get the right coverage information. And if coverage doesn't work on main, perhaps the commit that fixes coverage in this PR can be cherry-picked into a separate branch and submitted as a separate PR that can be merged first to establish the baseline coverage?

.github/workflows/main.yml
```diff
@@ -24,77 +24,83 @@ jobs:
...
-            image: ubuntu-24.04   # switch to Noble (24.04)
+            image: ubuntu-24.04
       fail-fast: false
     defaults:
       run:
-        shell: bash
+        shell: bash -l {0}
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v5
```
pjonsson: Is there a reason to not use version 6?

.github/workflows/main.yml
```diff
@@ -14,8 +14,8 @@ on:
   workflow_dispatch:
 
 env:
-  POETRY_VERSION: 1.5.1
-
+  POETRY_VERSION: 2.3.3
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true  # fixes the Node.js 20 warning
```
pjonsson: It's nice if the comment contains a note about which action that triggers "the Node.js 20 warning" so this variable can be removed when that action is updated in the future.

.pre-commit-config.yaml
```diff
@@ -1,38 +1,39 @@
...
+    rev: v3.21.2 # upgrading to 3.21.2 from 3.10.1 to accomodate py3.14 in test_matrix
     hooks:
       - id: pyupgrade
         args:
           - --keep-runtime-typing
           - --py39-plus
 
--   repo: https://github.com/python-poetry/poetry
+  - repo: https://github.com/python-poetry/poetry
     rev: 1.5.1
```
pjonsson: The `POETRY_VERSION` in the CI is set to `2.3.3` by this PR, should the version here be the same?

openeo_processes_dask/process_implementations/arrays.py
```diff
@@ -421,15 +427,20 @@ def array_interpolate_linear(data: ArrayLike, axis=None, dim_labels=None):
...
+        # - via np.apply_along_axis (axis path, already computed)
+        # - via da.map_blocks (1D dask path, one chunk at a time)
+        data = np.asarray(data)
         valid = np.isfinite(data)
-        if (valid == 1).all():
+        if valid.all():
             return data
         if len(x[valid]) < 2:
             return data
+        data = data.copy()  # avoid mutating the caller's array
```
pjonsson: This looks like it changes both behavior of the `interp` function and how expensive it is to call it.

openeo_processes_dask/process_implementations/ml/random_forest.py
```diff
@@ -126,7 +131,10 @@ def predict_random_forest(
     axis: int = -1,
     context: dict = None,
 ) -> RasterCube:
-    import xgboost as xgb
+    try:
+        from xgboost import dask as dxgb
+    except ImportError as e:
+        raise ImportError("xgboost[dask] is required for predict_random_forest.") from e
```
pjonsson: It looks like the intention is to replace the import error generated by the system, shouldn't this raise `from None` instead to avoid showing a bunch of extra stuff in the backtrace?

openeo_processes_dask/process_implementations/ml/random_forest.py
```diff
@@ -28,7 +28,12 @@ def fit_regr_random_forest(
     target_var: str = None,
     **kwargs,
 ) -> Booster:
-    import xgboost as xgb
+    try:
+        from xgboost import dask as dxgb
+    except ImportError as e:
+        raise ImportError(
```
pjonsson, line 34: It looks like the intention is to replace the import error generated by the system, shouldn't this raise `from None` instead to avoid showing a bunch of extra stuff in the backtrace?

openeo_processes_dask/process_implementations/utils.py
suriyahgit: Thanks @pjonsson, it's a good point. I simplified this while keeping separate scalar and dtype-bearing paths. np.dtype(obj).type is not safe for all current callers: it raises for values like 1, 1.0, and ndarray inputs, and string values can be interpreted as dtype aliases. Scalars now use np.dtype(type(obj)).type, while dtype-bearing objects use np.dtype(obj.dtype).type.

openeo_processes_dask/process_implementations/arrays.py
suriyahgit: agreed! that this needed clearer separation, this is not specifically a NumPy 2 issue; np.isnan is only defined for numeric-like values, so the old unconditional check was too broad. I updated the comment and kept non-numeric search values on the normal equality/not-found path.

.github/workflows/main.yml
suriyahgit: there was no reason to avoid version 6, so actions/checkout is now updated to v6.

.github/workflows/main.yml
suriyahgit: that makes sense. addressed this by removing the override instead of documenting it. The workflow now uses actions/checkout@v6 and setup-miniconda@v4, so FORCE_JAVASCRIPT_ACTIONS_TO_NODE24 is no longer needed.

.pre-commit-config.yaml
suriyahgit: nice catch, yes, it should match, addressed this. The poetry-check pre-commit hook now uses 2.3.3 to align with POETRY_VERSION in CI.

openeo_processes_dask/process_implementations/arrays.py
suriyahgit: agreed, it indeed makes things expensive at this point. addressed in the latest commit: e91ab28. The previous lazy Dask path could process chunks independently and change interpolation results at chunk boundaries. I changed the 1D Dask path to rechunk to one block before map_blocks, preserving the eager/global interpolation behavior while still delaying execution until compute time. I also added a multi-chunk regression test for this.

openeo_processes_dask/process_implementations/ml/random_forest.py
suriyahgit: thanks, agreed. addressed in e91ab28. This now raises the clearer ImportError with from None, so the original import traceback is suppressed.

openeo_processes_dask/process_implementations/ml/random_forest.py
suriyahgit: this now also raises with from None, matching the intended replacement error behavior.

suriyahgit:
> > This likely stems from missing/invalid coverage reports on recent `main` commits, not from the changes in this PR. We should trigger a fresh CI run on main (or merge this PR) to ensure coverage is uploaded with the updated workflow, so Codecov can establish a valid baseline for future comparisons.
> 
> Does code coverage work for main, but CodeCov was down when the last commit was merged to main? If so, merging #358 and rebasing this on top of latest main should get the right coverage information. And if coverage doesn't work on main, perhaps the commit that fixes coverage in this PR can be cherry-picked into a separate branch and submitted as a separate PR that can be merged first to establish the baseline coverage?

@pjonsson I did not make a branch-level code change for this because it appears to be a baseline/main coverage upload issue rather than something this PR branch can fully fix. After e91ab28, Codecov is comparing against the pushed head and patch coverage improved, but it still reports the base as 010c214 and 38 commits behind main. Rebasing onto a main commit with a valid coverage upload should make the comparison meaningful. 

@pjonsson thanks again for your time and efforts for reviewing this PR

I pushed e91ab28 to address the review feedback from @pjonsson . 
Main changes: updated the workflow action versions, removed the Node24 override, aligned the Poetry pre-commit hook with CI, simplified get_scalar_type, clarified array_find NaN handling, preserved global behavior for lazy 1D Dask interpolation, and switched the XGBoost import guards to raise from None.

PR #379, opened 2026-06-05
CI: pin actions by sha256

pjonsson:
Update the actions to the latest
version and pin them by sha256
to avoid supply-chain attacks
if the action provider gets
compromised.

### DIFF
--- a/.github/workflows/install-methods.yml
+++ b/.github/workflows/install-methods.yml
@@ -17,8 +17,8 @@ jobs:
       run:
         shell: bash -l {0}
     steps:
-      - uses: actions/checkout@v6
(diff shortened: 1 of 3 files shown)


pjonsson:
@suriyahgit since you just changed the CI, do you know why this is failing? (GDAL 3.13.1 was released 10 minutes ago.)

I rebased #364 to get a comparison, and the CI fails with that as well now, so the issue isn't related to this change.

suriyahgit:
Thanks @pjonsson for highlighting this, i agree to the latest actions versions and pin the actions versions by sha256 - as a sec fix.

I started to investigate a way on how to accomodate this latest gdal 3.13.1 which has just got released an hour ago.

PR #380, opened 2026-06-05
Pin GDAL to 3.13.0 for CI; keep library dep open-ended (>=3.8.4)

suriyahgit:
Pinning the ceiling version for gdal in the pyproject.toml and also the ci

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ packages = [
 
 [tool.poetry.dependencies]
 python = ">=3.10,<3.14"
-gdal = { version = ">=3.9", optional = true }
+gdal = { version = ">=3.8.4", optional = true }
 geopandas = { version = ">=0.11.1", optional = true }
 pandas = { version = ">=2.0.0", optional = true }
 xarray = { version = ">=2022.11.0", optional = true }
@@ -52,6 +52,11 @@ pystac = { version = ">=1.8.0", optional = false }
 zarr = "<=2.18.7"
 
 
+[tool.poetry.group.ci.dependencies]
+# CI-only version ceilings: protect CI from upstream breakage without
+# imposing ceilings on downstream consumers. See README for policy.
+gdal = { version = "<3.13.1" }
+
 [tool.poetry.group.dev.dependencies]
 pytest = "^7.2.0"
 ipykernel = "^6.15.1"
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -25,7 +25,7 @@ packages = [
 
 [tool.poetry.dependencies]
 python = ">=3.10,<3.14"
-gdal = { version = ">=3.9", optional = true }
+gdal = { version = ">=3.8.4,<3.13.1", optional = true }
```
pjonsson: I don't know the Poetry syntax for this, but my understanding is that it's the CI for this repository and/or Conda that can't deal with GDAL 3.13.1. Could the upper limit for gdal be left open, and an optional dependency along the lines of this be introduced into pyproject.toml:
```toml
[project.optional-dependencies]
civersions = ['gdal < 3.13.1']
```
and then the CI install that for testing?

This change would keep the package installable for people who use other "distributions" of GDAL, for example the GDAL Docker images.

(I can easily introduce a missing upper bound in a package that depends on openeo-processes-dask, but it's hard to remove the upper bound set by some other package.)

pjonsson:
I think this PR is one logical change, and it's 5 lines of diff so the change is small, I suggest you squash the commits together so `git bisect` can't land on one of the intermediate states where the CI still doesn't work.  Squashing the commits together would also make it possible to look at git blame for `README.md` and find the commit that introduced these bounds in `pyproject.toml`.

suriyahgit:
@pjonsson Thanks for acting as an guiding light. I would take this https://github.com/Open-EO/openeo-processes-dask/pull/380#issuecomment-4634451268 as a general comment for my future commits.

**_Both points are addressed:_**
1. Upper bound - Removed <3.13.1 from the main gdal dependency. Added [tool.poetry.group.ci.dependencies] with gdal <3.13.1 and a civersions extra for discoverability. CI now runs poetry install --with dev,ci --all-extras to activate the ceiling. The conda pin gdal=3.13.0 in ci-environment.yml remains as the exact execution pin.
2. Squash - Squashed everything into a single commit 13e956a.

pyproject.toml
suriyahgit: A very good idea, i tried implementing this with this PR.

pjonsson (approved):
I don't have any formal powers in openEO, but this will work for our use-case once we catch up to the latest release.

pjonsson:
> I would take this [#380 (comment)](https://github.com/Open-EO/openeo-processes-dask/pull/380#issuecomment-4634451268) as a general comment for my future commits.

That would be great, and thank you for the work on making the CI green again!

If you're interested in some elaborated reasoning for why I suggested the squashing, this blog post from 2009 is still a great summary of both some good and bad habits: https://who-t.blogspot.com/2009/12/on-commit-messages.html

suriyahgit:
>https://who-t.blogspot.com/2009/12/on-commit-messages.html

will definitely give a quick read! Thanks😊

PR #381, opened 2026-06-09
CI: add dependabot configuration

pjonsson:
This adds configuration for getting
automatic pull requests for version
updates for GitHub actions.

Updating CI actions rarely breaks,
so group all updates into a single
pull request. Also use a 10 day
cooldown to give some time for
people to discover and withdraw
any releases made by bad actors.

https://docs.github.com/en/code-security/dependabot


clausmichele:
Will the bot also pin the actions once updated as done here? https://github.com/Open-EO/openeo-processes-dask/pull/379

pjonsson:
Yeah, Dependabot will use pins if that is what is there previously.

Issue #383, opened 2026-06-11
specs shipped as SSH git submodule breaks installs from source/git


turban:
### Problem

The openEO process spec JSONs are pulled in as a **git submodule over SSH**, per `.gitmodules` on `main`:

```ini
[submodule "openeo_processes_dask/specs/openeo-processes"]
    path = openeo_processes_dask/specs/openeo-processes
    url = git@github.com:eodcgmbh/openeo-processes.git
```

This makes installing the package **from source / git** unreliable, because the specs directory is empty unless the submodule is separately initialised:

- `pip install "git+https://github.com/Open-EO/openeo-processes-dask.git"` does **not** fetch submodules, so the `specs/openeo-processes` tree is missing and spec lookups fail at runtime.
- An sdist / `git archive` checkout has the same gap unless cloned with `--recurse-submodules`.
- The `git@github.com:` (SSH) URL additionally requires an authenticated SSH key, so it fails in CI, containers, and any anonymous/HTTPS-only environment even when submodules *are* requested.

The published PyPI wheel is fine (the specs get bundled at build time), so this only bites consumers who install **unreleased fixes directly from the repo** — which is exactly what downstreams need to do today (e.g. to pick up the relaxed `requires-python <3.14` on `main` before it's released, or any future Zarr-related change from #376).

### Suggested fix

Vendor the spec JSON files directly into the package tree (drop the submodule), or at minimum switch the submodule URL to `https://` and ensure the build always bundles the JSONs. Vendoring is the most robust — it makes `pip install git+https://…` and sdist builds work with no extra steps.

### Context / offer

We maintain a fork where we did exactly this (detach the submodule, commit the JSON specs directly) to make `git+https` installs work for our deployment. Happy to open a PR upstreaming that change if it's a direction you'd accept.

pjonsson:
I'm also interested in getting a release of a version with the relaxed constraints. My last couple of PRs have been merged within a day or two, so maybe we can get a release out tomorrow.

My experience is that vendoring decisions tend to drag out on time (and it's not completely unjustified, vendoring brings additional maintenance burden), so I made a stop-gap measure PR that switches to using https. I guess we'll see if anyone objects to that.

clausmichele:
@turban version 2026.6.3 includes https for the processes specs, is this issue solved now?

turban:
[...]

PR #384, opened 2026-06-11
gitmodules: use https

pjonsson:
The SSH URL requires
an authenticated SSH key,
so it fails in containers
and other similar environments.
Switch to using https instead.

Refs #383


clausmichele:
This was already addressed in the past, strange that it appeared again, thanks.

PR #385, opened 2026-06-11
pyproject: relax zarr version

pjonsson:
It looks like it's odc-loader
that effectively constrains
the version of the zarr package,
so depend on odc-loader instead
of specifying our own constraints.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,10 @@ openeo-pg-parser-networkx = { version = ">=2025.10", optional = true }
 rqadeforestation = { version = ">=0.1", optional = true }
 odc-geo = { version = ">=0.4.1,<1", optional = true }
 stac_validator = { version = ">=3.3.1", optional = true }
+# odc-stac depends on odc-loader, but it's odc-loader that needs to be
+# compatible with the zarr package, so depend on odc-loader to get the right
+# version constraint for the zarr package.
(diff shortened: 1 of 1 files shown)


pjonsson:
This sits on top of #386, so that should be merged first.

suriyahgit:
Thanks @pjonsson for this — the `odc-loader[zarr]` approach is cleaner than maintaining our own `zarr` constraint. We've adopted the same dependency strategy in our `zarr-v3-compat-plan` branch at `Eurac-Research-Institute-for-EO/zarr-v3-compat-plan`.

However, we found that fixing just `pyproject.toml` isn't enough for full Zarr v3 support. At runtime, `load_stac()` still:

1. Forces `consolidated=True` — unsupported by `zarr-python 3`, causing an error when opening any v3 store.
2. Ignores `xarray:open_kwargs` from STAC assets — so v3 STAC items that specify `zarr_version: 3` via `xarray:open_kwargs` are silently dropped.
3. Doesn't map `zarr_version` → `zarr_format` — Xarray expects `zarr_format`, but STAC uses `zarr_version`.

We'd suggest reviewing this PR: https://github.com/Open-EO/openeo-processes-dask/pull/393  as the complete fix - it builds on your `odc-loader` approach with the same `pyproject.toml` changes, plus adds the `_zarr_open_kwargs` helper and 6 new tests covering v2/v3 scenarios. That way we address both the install constraint and the runtime compatibility in one go.

PR #388, opened 2026-06-13
pyproject: update to pytest 9

pjonsson:
Bump pytest and pytest-cov to
more recent versions. This
fixes CVE-2025-71176.

Also enable the "new" output
style "times" that prints
execution times for each
test.


pjonsson:
@clausmichele I have an identical PR on openeo-pg-parser-networkx, could you please take a look at that as well?

clausmichele:
@pjonsson I did see it, but I don't have rights on that repo yet.

@ValentinaHutter let me know if I can be added as maintainer there too.

PR #391, opened 2026-06-17
CI: limit concurrency on workflows

pjonsson:
Cancel any running CI jobs for
the previous iteration of
a pull request, but always
let the CI jobs for the main
branch run to completion.

The number of concurrent CI jobs
is limited for the organization,
so this ensures this repository
doesn't gobble up all the CI
runners when someone is repeatedly
updating their pull request.


clausmichele:
Thanks, another good contribution!

PR #392, opened 2026-06-17
pyproject: allow Python 3.14

pjonsson:


pjonsson:
Not sure why Python 3.14 was excluded earlier, it seems to pass the tests without any further changes.

suriyahgit:
@pjonsson fiona dependency crashed when we introduced py3.14, that's the reason of exclusion. Fiona currently lacks stable, official wheel support for Python 3.14

pjonsson:
@suriyahgit there isn't even a mention of fiona in this repository, so that's hard to find. Could you please add something to the test suite that resembles the thing that triggered the problem for you so we at least have some kind of safeguard against future regressions?

suriyahgit:
@pjonsson it comes from geopandas dependency, although it is optional. I ll try reproducing the error.

pjonsson:
@suriyahgit by all means, do it during regular work hours, I'm just commenting here when I happen to be working (I interact with Australia a fair bit).

pjonsson:
@clausmichele I've fixed the broken merge, so this should pass now.

clausmichele:
> @clausmichele I've fixed the broken merge, so this should pass now.

Ahah, you're too fast, I was also doing it!

Issue #376, opened 2026-05-28
Zarr v3 support — is it planned?


turban:
We're building a data backend on zarr v3 and icechunk, and would love to integrate openeo-processes-dask for process graph execution. However, the current `zarr = "<=2.18.7"` constraint makes it impossible to install alongside zarr >=3.

Is zarr v3 support on the roadmap?

turban:
Looking more closely, it seems zarr is never actually imported or used directly in the openeo-processes-dask source — all zarr operations go through xarray. The `zarr = "<=2.18.7"` constraint appears to be a vestigial direct dependency rather than something the library actually needs.

If that's correct, removing zarr from direct dependencies entirely would unblock zarr v3 users without requiring any code changes.

clausmichele:
Hi @turban, thanks for the feedback. We are currently reviewing the dependencies of the project and will soon fix Zarr too.

turban:
Following up with a concrete data point from a downstream that hit this.

In the current PyPI release (`openeo-processes-dask==2026.6.1`), the `zarr<=2.18.7` cap is a **base/core dependency** — it has no `; extra == ...` marker, unlike most of the heavy deps:

```
rasterio<2.0.0,>=1.3.4 ; extra == "implementations"
odc-geo<1,>=0.4.1      ; extra == "implementations"
...
zarr<=2.18.7           # no extra marker → installed unconditionally
```

So it can't be sidestepped by selecting/skipping an extra — `pip install openeo-processes-dask` pins Zarr 2.x for everyone. That makes the package incompatible with a Zarr v3 / icechunk data backend, which is our setup. We currently have to force `zarr>=3` via a resolver-level dependency override to use the library at all, which is fragile.

Is there anything in the implementations that actually requires Zarr 2.x at runtime, or is the cap just a stale upper bound? If it's the latter, relaxing it (or at least moving it behind the `implementations` extra) would unblock Zarr v3 downstreams without waiting on full v3 support. Happy to help test against Zarr v3 if useful.

pjonsson:
> Is there anything in the implementations that actually requires Zarr 2.x at runtime, or is the cap just a stale upper bound? If it's the latter, relaxing it (or at least moving it behind the `implementations` extra) would unblock Zarr v3 downstreams without waiting on full v3 support. Happy to help test against Zarr v3 if useful.

[...]

PR #393, opened 2026-06-17
Zarr v3 support

suriyahgit:
## Description

This PR adds runtime Zarr v3 support to `openeo-processes-dask`, addressing `#376` and building on the dependency-only fix from PR `#385`.

PR `#385` removed the hard `zarr = "<=2.18.7"` constraint and added `odc-loader[zarr]`, which resolved the install conflict. However, Zarr v3 stores could still fail at runtime because `load_stac()` hardcoded Zarr open options.

## Changes

* Adds `_zarr_open_kwargs()` in `load.py`

  * Respects STAC `xarray:open_kwargs`
  * Maps `zarr_version` to `zarr_format`
  * Avoids forcing `consolidated=True`, which is unsupported by `zarr-python 3`
  * Keeps sensible defaults: `engine="zarr"` and `chunks={}`

* Updates dependency handling in `pyproject.toml`

  * Replaces the direct Zarr ceiling with `odc-loader[zarr]`
  * Adds `xarray >=2025.06.0` to the CI group for Zarr v3 `_FillValue` compatibility
  * Adds `odc-loader` to implementation extras

* Adds 6 tests covering Zarr v2/v3 loading scenarios, open kwargs handling, missing band errors, and temporary STAC item helpers.

## Testing

```text
319 tests passed, 1 skipped
```

The skipped test is the Zarr v3 test when `zarr-python 3` is not installed.

### DIFF
--- a/openeo_processes_dask/process_implementations/cubes/load.py
+++ b/openeo_processes_dask/process_implementations/cubes/load.py
@@ -12,6 +12,7 @@ import numpy as np
 import odc.stac
 import planetary_computer as pc
 import pyproj
+import pystac
 import pystac_client
 import xarray as xr
 from openeo_pg_parser_networkx.pg_schema import BoundingBox, TemporalInterval
@@ -82,6 +83,19 @@ def _search_for_parent_catalog(url):
             "It was not possible to find the root STAC Catalog starting from the provided Collection."
         )
     return catalog_url, collection_id
+
+
+def _zarr_open_kwargs(asset: pystac.Asset, use_xarray_open_kwargs: bool) -> dict:
+    kwargs = (
+        dict(asset.extra_fields.get("xarray:open_kwargs", {}))
+        if use_xarray_open_kwargs
+        else {}
+    )
+    kwargs.setdefault("engine", "zarr")
+    kwargs.setdefault("chunks", {})
+    if "zarr_version" in kwargs and "zarr_format" not in kwargs:
+        kwargs["zarr_format"] = kwargs.pop("zarr_version")
+    return kwargs
 
 
 def load_stac(
@@ -239,11 +253,7 @@ def load_stac(
         datasets = []
         for item in items:
             for asset in item.assets.values():
-                kwargs = (
-                    asset.extra_fields.get("xarray:open_kwargs", {})
-                    if use_xarray_open_kwargs
-                    else {"engine": "zarr", "consolidated": True, "chunks": {}}
-                )
+                kwargs = _zarr_open_kwargs(asset, use_xarray_open_kwargs)
 
                 if use_xarray_storage_options:
                     storage_opts = asset.extra_fields.get("xarray:storage_options", {})
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,31 +25,34 @@ packages = [
 
 [tool.poetry.dependencies]
 python = ">=3.10,<3.14"
+dask = {extras = ["array", "dataframe", "distributed"], version = ">=2023.4.0", optional = true}
+dask-geopandas = { version = ">=0.4.3", optional = true }
 gdal = { version = ">=3.8.4", optional = true }
 geopandas = { version = ">=0.11.1", optional = true }
-pandas = { version = ">=2.2.0", optional = true }
-xarray = { version = ">=2022.11.0", optional = true }
-dask = {extras = ["array", "dataframe", "distributed"], version = ">=2023.4.0", optional = true}
-rasterio = { version = "^1.3.4", optional = true }
-dask-geopandas = { version = ">=0.4.3", optional = true }
-xgboost = { version = ">=1.5.1", extras = ["dask"], optional = true }
-rioxarray = { version = ">=0.12.0,<1", optional = true }
-openeo-pg-parser-networkx = { version = ">=2025.10", optional = true }
-rqadeforestation = { version = ">=0.1", optional = true }
-odc-geo = { version = ">=0.4.1,<1", optional = true }
-stac_validator = { version = ">=3.3.1", optional = true }
-odc-stac = { version = ">=0.3.9", optional = true }
-pystac_client = { version = ">=0.6.1", optional = true }
-planetary_computer = { version = ">=0.5.1", optional = true }
-scipy = "^1.11.3"
-xvec = { version = "0.2.0", optional = true }
-joblib = { version = ">=1.3.2", optional = true }
 geoparquet = "^0.0.3"
-pyarrow = ">=15.0.2"
-openeo = ">=0.36.0"
+joblib = { version = ">=1.3.2", optional = true }
 numpy = { version = ">=1.26.3,<3", optional = false }
+odc-geo = { version = ">=0.4.1,<1", optional = true }
+# odc-stac depends on odc-loader, but it's odc-loader that needs to be
+# compatible with the zarr package, so depend on odc-loader to get the right
+# version constraint for the zarr package.
+odc-loader = { version = ">=0.5", extras = ["zarr"], optional = true }
+odc-stac = { version = ">=0.3.9", optional = true }
+openeo = ">=0.36.0"
+openeo-pg-parser-networkx = { version = ">=2025.10", optional = true }
+pandas = { version = ">=2.2.0", optional = true }
+planetary_computer = { version = ">=0.5.1", optional = true }
+pyarrow = ">=15.0.2"
 pystac = { version = ">=1.12.0", optional = true }
-zarr = "<=2.18.7"
+pystac_client = { version = ">=0.6.1", optional = true }
+rasterio = { version = "^1.3.4", optional = true }
+rioxarray = { version = ">=0.12.0,<1", optional = true }
+rqadeforestation = { version = ">=0.1", optional = true }
+scipy = "^1.11.3"
+stac_validator = { version = ">=3.3.1", optional = true }
+xarray = { version = ">=2022.11.0", optional = true }
+xgboost = { version = ">=1.5.1", extras = ["dask"], optional = true }
+xvec = { version = "0.2.0", optional = true }
 
 
 [tool.poetry.group.ci.dependencies]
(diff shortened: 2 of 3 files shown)


turban:
Thanks @suriyahgit — this looks great and addresses #376 well.

Dropping the forced `consolidated=True` is the key runtime fix (zarr-python 3 doesn't support it), and mapping `zarr_version` → `zarr_format` while respecting `xarray:open_kwargs` is a nice back-compat touch. Relaxing the constraint via `odc-loader[zarr]` instead of the hard `zarr <=2.18.7` ceiling is exactly what we needed on our zarr v3 + icechunk backend.

LGTM from our side — thanks for picking this up.

openeo_processes_dask/process_implementations/cubes/load.py
```diff
@@ -238,11 +251,7 @@ def load_stac(
         datasets = []
         for item in items:
             for asset in item.assets.values():
-                kwargs = (
-                    asset.extra_fields.get("xarray:open_kwargs", {})
-                    if use_xarray_open_kwargs
-                    else {"engine": "zarr", "consolidated": True, "chunks": {}}
-                )
+                kwargs = _zarr_open_kwargs(asset, use_xarray_open_kwargs)
```
pjonsson, line 256: Can't the first argument just be `asset.extra_fields.get("xarray:open_kwargs", {}) if use_xarray_open_kwargs else {}` and the second argument is no longer required?

pyproject.toml
```diff
@@ -49,13 +49,18 @@ pyarrow = ">=15.0.2"
 openeo = ">=0.36.0"
 numpy = { version = ">=1.26.3,<3", optional = false }
 pystac = { version = ">=1.12.0", optional = true }
-zarr = "<=2.18.7"
+# odc-stac depends on odc-loader, but it's odc-loader that needs to be
+# compatible with the zarr package, so depend on odc-loader to get the right
+# version constraint for the zarr package.
+odc-loader = { version = ">=0.5", extras = ["zarr"], optional = true }
```
pjonsson, line 39: I can see the dependencies aren't sorted in alphabetical order, but I don't understand the way they are currently grouped either. The reason I put odc-loader on the line before odc-stac is because odc-loader is a dependency of odc-stac, so it makes sense to keep them close to each other, and that also happens to be sorted alphabetically.

tests/test_load_stac.py
```diff
@@ -1,18 +1,79 @@
...
-@pytest.mark.parametrize("size", [(10, 10, 10, 5)])
-@pytest.mark.parametrize("dtype", [np.float32])
-# @pytest.mark.parametrize("dims", [("x", "y", "t", "bands")])
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _can_write_zarr_v3():
```
pjonsson: Just catching every exception and claiming the setup can't write Zarr v3 will make every broken setup skip the test, so regressions won't be detected? Shouldn't this test if the Zarr v3 dependency is installed?
```python
from importlib.metadata import version
from packaging.version import parse as parse_version

zarr_v3 = parse_version(version("zarr")) >= parse_version("3")
```

tests/test_load_stac.py
```diff
@@ -1,18 +1,79 @@
...
+    try:
+        ds = xr.Dataset({"a": (("x",), [1])})
+        with tempfile.TemporaryDirectory() as tmp:
+            ds.to_zarr(pathlib.Path(tmp) / "v3.zarr", zarr_format=3)
+        return True
+    except Exception:
```
pjonsson: I'm guessing this function can be replaced with the `zarr_v3` variable, but only the `ds.to_zarr()` should be within the try-block so errors in other parts aren't detected as Zarr v3 issues, and `except Exception` is so broad that it will catch a bunch of other errors that aren't related to Zarr v3.

tests/test_load_stac.py
```diff
@@ -1,18 +1,79 @@
...
+    *,
+    open_kwargs=None,
+    storage_options=None,
+) -> pathlib.Path:
+    """Create a temporary STAC Item JSON pointing to *zarr_path*.
+
+    If *open_kwargs* is ``None`` the ``xarray:open_kwargs`` field is
+    removed from the asset.  Same for *storage_options*.
+    """
+    with open(_STAC_TEMPLATE_PATH) as f:
```
pjonsson: I feel the STAC adjustment should use pystac since we already depend on it. (And `pystac.read_file(_STAC_TEMPLATE_PATH)` followed by some adjustments followed by `pystac.write_file(item, dest_href=stac_path)` doesn't require this repository to worry about ensuring standard compliance for STAC files at least.)

tests/test_load_stac.py
```diff
@@ -51,6 +112,102 @@ def test_load_stac(bounding_box, random_raster_data, temporal_interval):
...
+
+@pytest.mark.parametrize("size", [(10, 10, 10, 5)])
+@pytest.mark.parametrize("dtype", [np.float32])
+def test_load_stac_zarr_v2_unconsolidated_without_open_kwargs(
+    bounding_box, random_raster_data, temporal_interval, tmp_path
+):
+    cube = _create_cube_for_zarr(bounding_box, random_raster_data, temporal_interval)
+    zarr_path = tmp_path / "v2_unconsolidated.zarr"
+    cube.to_dataset(dim="bands").to_zarr(
+        str(zarr_path), zarr_format=2, consolidated=False
```
pjonsson: From what I can tell, this test is identical to the next test except the consolidated value. Maintaining tests also costs time, so make it a single test that also is also parameterized on the consolidated value.

tests/test_load_stac.py
```diff
@@ -51,6 +112,102 @@ def test_load_stac(bounding_box, random_raster_data, temporal_interval):
...
+@pytest.mark.skipif(
+    not _can_write_zarr_v3(),
+    reason="Zarr-Python 3 not available in this environment",
+)
+@pytest.mark.parametrize("size", [(10, 10, 10, 5)])
+@pytest.mark.parametrize("dtype", [np.float32])
+def test_load_stac_zarr_v3_with_open_kwargs(
+    bounding_box, random_raster_data, temporal_interval, tmp_path
+):
+    cube = _create_cube_for_zarr(bounding_box, random_raster_data, temporal_interval)
```
pjonsson: I'm guessing this test is identical to the first test. Combine the two tests and parameterize on the zarr version, and add something along the lines of this first in the test body:
```python
if zarr_version == 3 and parse_version(version("zarr")) < parse_version("3"):
    pytest.skip("Zarr v3 is not supported in this configuration")
```

pjonsson:
I realize the type signatures for old code leave some room for improvement, but it's quite a lot of work to go through and figure out what people were thinking in the past and potentially make corrections. But it's usually quite easy to add type annotations when adding new functions, so it would be great if you added type annotations for those.

I don't know how releases work in this repository, but the PR title usually tends to end up in the release notes one way or another in my experience. In this case, "Add Zarr v3 support" seems like a suitable title since that is the "feature" this brings, and implementation details like helpers/dependency changes/tests can be mentioned in the PR description instead (if they are important).

openeo_processes_dask/process_implementations/cubes/load.py
suriyahgit: @pjonsson the use_xarray_open_kwargs flag is there on purpose. The xarray:open_kwargs field comes from the deprecated xarray-assets (https://github.com/stac-extensions/xarray-assets) extension, so the flag makes it obvious when we're opting into that old behavior rather than silently picking it up from extra_fields. It also keeps the call site self-documenting.

pyproject.toml
suriyahgit: @pjonsson sorted all dependencies alphabetically and kept odc-loader next to odc-stac with the comment. Good one.

tests/test_load_stac.py
suriyahgit: thanks for the suggestions, accomodated: replaced with importlib.metadata check: _zarr_v3_available() now just checks the installed zarr version instead of catching every exception.

tests/test_load_stac.py
suriyahgit: @pjonsson since we now use the package version check instead of a trial write, there's no try-block at all for the skip logic. The Zarr v3 skip condition just checks the installed version.

tests/test_load_stac.py
suriyahgit: @pjonsson combined into test_load_stac_zarr_v2_without_open_kwargs with @pytest.mark.parametrize("consolidated", [True, False]).

tests/test_load_stac.py
suriyahgit: @pjonsson combined into test_load_stac_zarr_with_open_kwargs with @pytest.mark.parametrize("zarr_version", [2, 3]) and a body-level skip when zarr v3 isn't available.

tests/test_load_stac.py
suriyahgit: @pjonsson switched _make_zarr_stac_item() to pystac.read_file() / pystac.write_file() instead of raw JSON dict manipulation. Thanks for the suggestion.

tests/test_load_stac.py
```diff
@@ -1,18 +1,75 @@
...
-# @pytest.mark.parametrize("dims", [("x", "y", "t", "bands")])
 
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _zarr_v3_available() -> bool:
+    try:
+        return parse_version(_pkg_version("zarr")) >= parse_version("3")
```
pjonsson: This shouldn't raise an exception under any reasonable circumstances so the try-statement can be removed. The statement will also evaluate to the same thing each time it's called, so that was why I suggested a global variable in the this test file instead (`_zarr_v3_available = parse_version(_pkg_version("zarr")) >= parse_version("3")`).

suriyahgit:
> I realize the type signatures for old code leave some room for improvement, but it's quite a lot of work to go through and figure out what people were thinking in the past and potentially make corrections. But it's usually quite easy to add type annotations when adding new functions, so it would be great if you added type annotations for those.
> 
> I don't know how releases work in this repository, but the PR title usually tends to end up in the release notes one way or another in my experience. In this case, "Add Zarr v3 support" seems like a suitable title since that is the "feature" this brings, and implementation details like helpers/dependency changes/tests can be mentioned in the PR description instead (if they are important).

great comment, atleast for this PR i have made sure to type annotate now.

tests/test_load_stac.py
suriyahgit: @pjonsson _zarr_v3_available is now available as a module-level bool constant and dropped the try/except

pjonsson (approved):
I don't know anything about Zarr, but I think the code looks fine.

PR #396, opened 2026-06-21
text: fix type annotations

pjonsson:
It's sufficient to use a boolean
for case_sensitive, and the functions
return bool rather than string.
Also switch to using Python 3.10
types when changing these lines.


pjonsson:
Intermittent error in the CI causing the test failure, but I don't have permissions to re-start the job.

clausmichele (approved):

PR #404, opened 2026-06-26
Import from right module

pjonsson:
These symbols are not exported
from where they are imported,
so update the code to import
from the right place.

### DIFF
--- a/openeo_processes_dask/process_implementations/math.py
+++ b/openeo_processes_dask/process_implementations/math.py
@@ -3,6 +3,7 @@ import logging
 import dask
 import dask.array as da
 import numpy as np
+from dask.delayed import delayed
 
 from openeo_processes_dask.process_implementations.cubes.utils import (
     _has_dask,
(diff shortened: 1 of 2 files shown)


openeo_processes_dask/process_implementations/math.py
```diff
@@ -3,6 +3,7 @@
 import dask
 import dask.array as da
 import numpy as np
+from dask.delayed import delayed
```
clausmichele, line 6: Shouldn't it be: `from dask import delayed` ? See https://github.com/dask/dask/blob/e237122a298a00b6091712d83d77c00e94bf30fc/dask/delayed.py#L460

pjonsson: Well, the Pyright author disagrees about exports: https://github.com/microsoft/pyright/issues/2639#issuecomment-983098575

I haven't read all the relevant standards so I can't comment on this issue, but my experience is that he's usually right on a technical level (and he tends to be balanced, so he's working within the system when some specification is saying crazy things).

From a practical perspective, Pylance is using Pyright under the hood, so every VSCode user will get squiggly lines under these symbols.

PR #407, opened 2026-06-29
pyproject: update to ipykernel 7.x

pjonsson:
Things still appear to work as
they are with version 7, so remove
the upper bound.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,7 +73,7 @@ ci = [
 ]
 dev = [
     "folium >=0.12.1,<1",
-    "ipykernel >=6.15.1, <7",
+    "ipykernel >=6.15.1",
     "mapclassify >=2.4.3, <3",
     "matplotlib >=3.5.3, <4",
     "pre-commit >=2.20.0, <3",


clausmichele:
So, I wonder if the following dev dependencies should even be mentioned at all:
```
    "folium >=0.12.1,<1",
    "ipykernel >=6.15.1, <7",
    "mapclassify >=2.4.3, <3",
    "matplotlib >=3.5.3, <4", 
```
In my opinion, if someone wants to test/develop this package in an interactive environment, where those packages are necessary, they should install them separately. What do you think @suriyahgit ?

pjonsson:
> In my opinion, if someone wants to test/develop this package in an interactive environment, where those packages are necessary, they should install them separately. What do you think @suriyahgit ?

I don't use those myself, but carrying those dependencies isn't terribly expensive (at least not if we lift the upper bound so we don't have to bump them regularly). Perhaps there should be a `dev-interactive` group or similar for those dependencies if they're not required for running tests/linting/etc.?

pjonsson:
@clausmichele I think this PR is a strict improvement over the current main branch (and it doesn't make things worse regardless of which way we choose to go in the future), could you please merge this for now and we can discuss the presence of non-essential dev-dependencies in some issue instead?

Besides my PR that removes unused imports, all my remaining PRs should also be harmless. I don't believe that removing the unused imports should break anything, but some import could be there for side effects without me realizing that is why it's there.

Issue #330, opened 2025-07-07
dimension names not passed in UDFs


clausmichele:
In the [current implementation](https://github.com/Open-EO/openeo-processes-dask/blob/34c04f23ed05310e8e7c8314f387b02caa063421/openeo_processes_dask/process_implementations/udf/udf.py#L14), which is relying on the one included in the openeo package here: https://github.com/Open-EO/openeo-python-client/blob/master/openeo/udf/run_code.py

The current implementation doesn't pass the dimension names (it renames them to "dim_0", "dim_1", ...) nor the dimension labels (converts all of them to integer). It would be necessary to have access to dimension names and labels inside the UDF.

Sample code to reproduce the issue:

```python
import openeo

from openeo.local import LocalConnection
local_conn = LocalConnection("./")

url = "https://stac.eurac.edu/collections/SENTINEL2_L2A_SAMPLE"

temporal_extent = ["2022-06-01T00:00:00Z", "2022-06-30T00:00:00Z"]

datacube = local_conn.load_stac(
    url=url,
    temporal_extent=temporal_extent,
    bands=["B03","B04"]
)

# Build a UDF object from an inline string with Python source code.
udf = openeo.UDF(
"""
import xarray
import numpy as np

def apply_datacube(cube: xarray.DataArray, context: dict) -> xarray.DataArray:
    print(cube.dims)
    return cube
"""
)

# Apply the UDF to a cube.
datacube.apply_dimension(dimension="bands",process=udf).execute()
```

`('dim_0', 'dim_1', 'dim_2', 'dim_3')`

@jzvolensky @Yuvraj198920 this would be a nice thing to fix within EOEPCA+.

PR #423, opened 2026-08-27
Fix/21 udf   dimensions names not passed to udf

elmaic:
not passing dimensions to UDF, added 1 file

solved here

https://github.com/Open-EO/openeo-processes-dask/issues/330


pjonsson:
This is a 76 commit +2000-300 line change, are you sure all these commits/changes are required to fix passing the dimension names to udf?

I also find some of the commits questionable, as an example, commit 0451640 sets the package version to a faulty value, and commit 052fd74 immediately corrects this. I understand these kind of mistakes happen during development, but why would we want to cement those two commits on the main branch of this repository? Same comment for the lint fixes, fixup'd to the commit they are correcting.

Orthogonal to the all the commits, I think the package release of the next version should be done by a separate pull request that isn't a 2000+ line change.

PR #428, opened 2026-08-31
fixes NDVI with target_band

### LINKED ISSUE #362: [bug] NDVI process, using target_band
Three seems to be a bug in the `NDVI` process. When using the `target_band` parameter, the process fails to merge the `data` datacube with the cube containing the NDVI cube.

The issue seems to be this line: https://github.com/Open-EO/openeo-processes-dask/blob/a7eb91fa2b3a514dd81c471ef2b93d4355dcdd46/openeo_processes_dask/process_implementations/cubes/indices.py#L57

`xr.merge` seems to be the wrong function to use here, as this function builds and returns a `xr.DataSet`, not a `xr.DataArray`, leading to this error, as the DataArrays are unnamed:

```
ValueError: unable to convert unnamed DataArray to a Dataset without providing an explicit name
```

If I am not mistaken, the correct way to do things would be

```python
nd = xr.concat([data, nd], dim=band_dim)
```

jonas-hurst:
fixes https://github.com/Open-EO/openeo-processes-dask/issues/362

- works when data.name is unset
- always returns an xr.DataArray (prev. a DataArray was returned when target_band was unset, and DataSet when it was set)
- when the original datacube has a secondary coordinate on the bands dimension, the NDVI band will have a default dummy value for this secondary coordinate

### DIFF
--- a/openeo_processes_dask/process_implementations/cubes/indices.py
+++ b/openeo_processes_dask/process_implementations/cubes/indices.py
@@ -64,10 +64,7 @@ def _add_missing_coords(
         else:
             fill = _dummy_value(coord.dtype)
 
-        missing[name] = (
-            dim,
-            np.full(target.sizes[dim], fill, dtype=reference.coords[name].dtype),
-        )
+        missing[name] = (dim, np.array([fill]))
 
     return target.assign_coords(**missing) if missing else target, reference
 
--- a/openeo_processes_dask/process_implementations/cubes/indices.py
+++ b/openeo_processes_dask/process_implementations/cubes/indices.py
@@ -44,24 +44,33 @@ def _add_missing_coords(
         DataArray `reference` with potentially altered coordinate dtype
     """
     missing = {}
-    for name, coord in reference.coords.items():
-        if name in target.coords:
-            continue
 
-        fill = _dummy_value(coord.dtype)
+    # get all secondary coordinates to dim in reference DataArray
+    secondary_coord_to_dim = [
+        (name, coord)
+        for name, coord in reference.coords.items()
+        if name != dim and coord.dims == (dim,)
+    ]
 
-        if coord.dims == ():  # scalar coordinate
-            missing[name] = ((), np.array(fill, dtype=coord.dtype))
+    for name, coord in secondary_coord_to_dim:
+        # special case: dim is "common_name"
+        if coord.dtype.kind == "U" and name == "common_name":
+            fill = "NDVI"
 
-        elif coord.dims == (dim,):  # coordinate along concat dim
-            missing[name] = (dim, np.full(target.sizes[dim], fill, dtype=coord.dtype))
+            # convert coord dtype to U<4 to fit "NDVI", if it is less than 4 characters
+            if coord.dtype.kind == "U" and int(coord.dtype.str[2:]) < 4:
+                reference = reference.assign_coords(common_name=coord.astype("<U4"))
 
-        elif all(d in target.dims for d in coord.dims):  # other dims
-            shape = tuple(target.sizes[d] for d in coord.dims)
-            missing[name] = (coord.dims, np.full(shape, fill, dtype=coord.dtype))
-        # else: dims don't exist in target -> can't sensibly create it, skip
+        else:
+            fill = _dummy_value(coord.dtype)
 
-    return target.assign_coords(**missing) if missing else target
+        missing[name] = (
+            dim,
+            np.full(target.sizes[dim], fill, dtype=reference.coords[name].dtype),
+        )
+
+    return target.assign_coords(**missing) if missing else target, reference
+
 
 def ndvi(
     data: RasterCube, nir: str = "nir", red: str = "red", target_band: str | None = None
@@ -109,7 +118,7 @@ def ndvi(
         nd = nd.expand_dims(band_dim).assign_coords({band_dim: [target_band]})
 
         # add potentially missing coords from data to nd, so that xr.concat works
-        nd = _add_missing_coords(nd, data, band_dim)
+        nd, data = _add_missing_coords(nd, data, band_dim)
         nd = xr.concat([data, nd], dim=band_dim)
 
     nd.attrs = data.attrs
--- a/openeo_processes_dask/process_implementations/cubes/indices.py
+++ b/openeo_processes_dask/process_implementations/cubes/indices.py
@@ -1,3 +1,4 @@
+import numpy as np
 import xarray as xr
 
 from openeo_processes_dask.process_implementations.data_model import RasterCube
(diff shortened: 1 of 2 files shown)


openeo_processes_dask/process_implementations/cubes/indices.py
```diff
@@ -12,6 +13,47 @@
...
+
+        if coord.dims == ():  # scalar coordinate
+            missing[name] = ((), np.array(fill, dtype=coord.dtype))
+
+        elif coord.dims == (dim,):  # coordinate along concat dim
+            missing[name] = (dim, np.full(target.sizes[dim], fill, dtype=coord.dtype))
+
+        elif all(d in target.dims for d in coord.dims):  # other dims
+            shape = tuple(target.sizes[d] for d in coord.dims)
+            missing[name] = (coord.dims, np.full(shape, fill, dtype=coord.dtype))
```
pjonsson: The fill values are created based on the type, so I assume the concrete values are never used. Can't this function use `np.empty()` instead to save the initialization cost?

openeo_processes_dask/process_implementations/cubes/indices.py
```diff
@@ -12,6 +13,47 @@
...
+    if np.issubdtype(dtype, np.timedelta64):
+        return np.timedelta64("NaT")
+    if np.issubdtype(dtype, np.integer):
+        return -1  # ints can't hold NaN
+    if np.issubdtype(dtype, np.bool_):
+        return False
+    return ""  # str / object
+
+
+def _add_missing_coords(target, reference, dim):
```
pjonsson: Even if some of the old code is missing type annotations, I think it's nice if new code where we know the intention provides type annotations.

openeo_processes_dask/process_implementations/cubes/indices.py
```diff
@@ -12,6 +13,47 @@
...
+    Missing coordinates are created with a dummy value
+    """
+    missing = {}
+    for name, coord in reference.coords.items():
+        if name in target.coords:
+            continue
+
+        fill = _dummy_value(coord.dtype)
+
+        if coord.dims == ():  # scalar coordinate
```
pjonsson: There doesn't seem to be any test coverage for this case with scalar coordinate and the case with "other dims". It's easy to break corner cases when making seemingly innocent changes, so please add something in the tests for those two cases.

openeo_processes_dask/process_implementations/cubes/indices.py
```diff
@@ -54,6 +96,10 @@ def ndvi(data: RasterCube, nir="nir", red="red", target_band=None):
         if target_band in data.coords:
             raise BandExists("A band with the specified target name exists.")
         nd = nd.expand_dims(band_dim).assign_coords({band_dim: [target_band]})
-        nd = xr.merge([data, nd])
+
+        # add potentially missing coords from data to nd, so that xr.concat works
+        nd = _add_missing_coords(nd, data, band_dim)
```
pjonsson: There are 4 consecutive assignments to a variable called "nd" (which I assume is short for normalized difference), and it seems unlikely that all 4 are the normalized difference.

If you rewrite this as:
```python
thing = _add_missing_coords(nd.expand_dims(band_dim).assign_coords({band_dim: [target_band]}), data, band_dim)
nd = xr.concat([data, thing], dim=band_dim)
```
we can reduce that to 2 consecutive assignments, and the surrounding code still works.

(I don't know what "thing" should be called instead, but you probably know what it is and can name it properly since you're in the process of modifying this code.)

openeo_processes_dask/process_implementations/cubes/indices.py
jonas-hurst: agreed

openeo_processes_dask/process_implementations/cubes/indices.py
jonas-hurst: I simplified the function and now only consider the (dim,) case, as all others are impossible to happenI simplified the function and now only consider the (dim,) case, as all others are impossible to happen

openeo_processes_dask/process_implementations/cubes/indices.py
jonas-hurst: I simplified the function and now only consider the (dim,) case, as all others are impossible to happen

openeo_processes_dask/process_implementations/cubes/indices.py
jonas-hurst: Disagree. Normalized difference is first computed, then dims and coords are changed to match the original datacube to which it will be added.
Also, your proposed change limits readability.

jonas-hurst:
Thanks for your quick review. I addressed your comments in my latest commits!

openeo_processes_dask/process_implementations/cubes/indices.py
```diff
@@ -12,7 +15,66 @@
...
+            # convert coord dtype to U<4 to fit "NDVI", if it is less than 4 characters
+            if coord.dtype.kind == "U" and int(coord.dtype.str[2:]) < 4:
+                reference = reference.assign_coords(common_name=coord.astype("<U4"))
+
+        else:
+            fill = _dummy_value(coord.dtype)
+
+        missing[name] = (
+            dim,
+            np.full(target.sizes[dim], fill, dtype=reference.coords[name].dtype),
```
pjonsson: The fill values are created based on the type, so I assume the concrete values are never used. Can't this function use `np.empty()` instead to avoid the cost of having to fill an entire array with dummy values?

openeo_processes_dask/process_implementations/cubes/indices.py
```diff
@@ -12,7 +15,66 @@
...
+    if np.issubdtype(dtype, np.bool_):
+        return False
+    return ""  # str / object
+
+
+def _add_missing_coords(
+    target: xr.DataArray, reference: xr.DataArray, dim: str
+) -> tuple[xr.DataArray, xr.DataArray]:
+    """
+    Adds missing secondary coordinate to dimension dim in 'reference' to 'target
```
pjonsson: Missing close quote on "target".

jonas-hurst: done.

openeo_processes_dask/process_implementations/cubes/indices.py
jonas-hurst: the array is always of length 1. I therefore simplified this to `np.array([fill])`

jonas-hurst: I disagree with the use use of np.empty(). The arbitrary random data in the resulting array may be misinterpreted by users or downstream openEO processes

pjonsson: If it's an array of length 1 it doesn't take a lot of time to initialize the array.

I don't think the fill values are any less arbitrary, but I guess they're deterministic and that is a valuable trait in itself.

clausmichele:
Thanks @jonas-hurst and @pjonsson for the contribution!

